PR #12, opened 2022-07-21
Documentation

elena-krismer:


elena-krismer:
The PR starts from the commit "color scheme" [1794f6c](https://github.com/MannLabs/alphastats/pull/12/commits/1794f6c8704c3b2cc06a226f713c8e22ba67569f)
The other commits are already reviewed I just messed around with the branches

settings.ini
```diff
@@ -0,0 +1,31 @@
...
+min_python = 3.6
+max_python = 4
+audience = Developers
+language = English
+custom_sidebar = False
+license = MIT
+status = 2
+nbs_path = nbs
+doc_path = docs
+doc_host = https://mannlabs.github.io/alphapept
```
straussmaximilian, line 20: Still a lot of alphapept in here.

straussmaximilian (approved):
Just realized how to restrict the commit range after reviewing so I checked all changes :)

### General comments:

- There are a couple of seemingly unused files. 
- For the documentation, I did not check the htmls and only skimmed over this. This will probably be easier to go over once we compile a PDF or so.

Some of the comments might seem pedantic but this helps to sell later.

As a next step we should probably test the code, e.g. if the installation and imports work.

PR #35, opened 2022-08-31
User Interface

elena-krismer:


alphastats/gui/pages/02_⚫ Import Data.py
```diff
@@ -0,0 +1,167 @@
...
+            # create dataset
+            st.session_state["dataset"] = DataSet(
+                loader=st.session_state.loader,
+                metadata_path=metadatafile_df,
+                sample_column=st.session_state.sample_column,
+            )
+            st.session_state["metadata_columns"] = metadatafile_df.columns.to_list()
+
+    if st.button("Create a DataSet without metadata"):
+        st.session_state["dataset"] = DataSet(loader=st.session_state.loader)
```
straussmaximilian, line 123: Here, when clicking, the user has no feedback on what has happened.

alphastats/gui/pages/02_⚫ Import Data.py
```diff
@@ -0,0 +1,167 @@
...
+        upload_metadatafile()
+
+
+def display_loaded_dataset():
+    st.info("Data was successfully imported")
+    st.info("DataSet has been created")
+    st.markdown(f"Raw data from {st.session_state.dataset.software}")
+    display_file(st.session_state.dataset.rawdata)
+    st.markdown(f"Metadata")
+    display_file(st.session_state.dataset.metadata)
```
straussmaximilian, line 149: This will cause an error if no metadata was loaded.

alphastats/gui/pages/03_⚫ Preprocessing.py
```diff
@@ -0,0 +1,52 @@
...
+        submitted = st.form_submit_button("Submit")
+
+    if submitted:
+        st.session_state.dataset.preprocess(
+            remove_contaminations=remove_contaminations,
+            subset=subset,
+            normalization=normalization,
+            imputation=imputation,
+        )
+        st.write("Data processed.")
```
straussmaximilian, line 37: `Data processed` has little information. How about redirecting the log here or displaying a summary of what was done?

straussmaximilian (changes requested):
Hi,

overall a lot of good changes. There was a lot of code, so a smaller PR would help to review it.
I used the MaxQuant testfile and selected the default values for preprocessing.


Comments:
The current entry point is not clear. 
- [x] if you pip install alphastas it will be in the environment and you would need to find out the path to launch it via `streamlit run`. You could use the CLI as before or write a custom entrypoint.

There are some bugs when no metadata is loaded:
- [x] e.g. Switching back to Import causes an Error
- [x]  Clicking Preprocess or Visualization causes an error

If you click on "Import dataset" and had one loaded previously, the upload dialog shows before selecting a software. When you upload a file then it causes an error. 

Suggestions:
- [x] We could put links to sample files so people can try with example data.
- [x] Uploading is not fool proof - e.g. I can select the metadata.xlsx as MaxQuant file and it will be previewed. An indicative error message can be helpful. 
- [x]  Metadata selection can be made fool proof, e.g. by checking if the columns actually have values that exists in the experimental data file. 
- [ ] Functions that are executed should potentially be disabled afterwards. E.g. when i repeatedly changed the sample column and created a dataset this caused an error.
- [x]  For the plots (e.g. PCA) they do not show the FileName and it is hard to know which datapoint belongs to what
- [ ] You could think of auto-rendering the plots instead of submitting them and e.g. use a spinner (see st.spinner) to indicate that work is going on.
- [ ] Plots: Color according to does not work if you choose the column you used as an identifier (e.g. samples - probably pandas set_index somewhere)
- [x]  Intensity-Plot has a bug: TypeError: plot_intensity() got an unexpected keyword argument 'color'
- [x]  Volcano-Plot: A multiselect for groups would be beneficial, e.g. if you want to compare individual files. For the Volcano plot a vertical line at zero could be helpfule to orient
- [ ]  I got an error selecting cluster map and dendrogram: ValueError: The condensed distance matrix must contain only finite values.
- [ ]  Statistical Analysis: Selecting the same group with default settings caused an error

Note: When using Emojis, they will be different on each OS, so things could look different on a windows system vs a mac system. 

I didn't test all options, we can also have a live session clicking through to spot some potential bugs etc.

alphastats/DataSet_Plot.py
```diff
@@ -125,7 +129,20 @@ def _plot_dimensionality_reduction(self, group, method, circle, **kwargs):
...
+        components["sample"] = sample_names
+
+        fig = px.scatter(components, x=0, y=1, labels=labels, color=group_color, hover_data=[components["sample"]])
+
+        # rename hover_data_0 to sample
+        fig_dict = fig.to_plotly_json()
+        data = fig_dict.get("data")
+        
+        for count, d in enumerate(data):
+            hover = d.get("hovertemplate").replace("hover_data_0", "sample")
```
straussmaximilian, line 142: This looks a bit like a hack.

You should be able to update the hover-text like so:

```
fig.update_traces(
    hovertemplate="<br>".join([
        "ColX: %{x}",
        "ColY: %{y}",
        "Col1: %{customdata[0]}",
        "Col2: %{customdata[1]}",
        "Col3: %{customdata[2]}",
    ])
)
```
See also: 
https://stackoverflow.com/questions/59057881/python-plotly-how-to-customize-hover-template-on-with-what-information-to-show

alphastats/DataSet_Statistics.py
```diff
@@ -35,6 +53,15 @@ def perform_diff_expression_analysis(self, column, group1, group2):
             * ``'coef_sd'``: the standard deviation of the coefficient in liker-space
             * ``'ll'``: the log-likelihood of the estimation
         """
+
+        import diffxpy.api as de
```
straussmaximilian, line 57: If this takes a long time a logging output can be useful.

alphastats/gui/gui.py
```diff
@@ -0,0 +1,7 @@
+import os
+
+
+def run():
+    file_path = os.path.realpath(__file__)
+    os.chdir(os.path.dirname(file_path))
+    os.system("streamlit run AlphaStats.py")
```
straussmaximilian, line 7: This could cause issues because it is not clear which streamlit installation is used.

alphastats/gui/pages/04_⚫ Visualization.py
```diff
@@ -4,8 +4,12 @@
 
 
 def display_plotly_figure(plot):
+    try:
+        st.plotly_chart(plot.update_layout(plot_bgcolor = "white"))
+    except:
```
straussmaximilian, line 9: Rare excepts are not recommended.
Maybe catch the exception and print?
e.g.

```
except Exception as e:
    print(e)
```

alphastats/gui/pages/04_⚫ Visualization.py
```diff
@@ -26,37 +30,34 @@ def make_plot():
...
 sidebar_info()
 
+
+# set background to white so downloaded pngs dont have grey background
+styl = f"""
+    <style>
+        .css-jc5rf5 {{
+            position: absolute;
+            background: rgb(255, 255, 255);
+            color: rgb(48, 46, 48);
```
straussmaximilian, line 48: Ideally have all colors in one central file,
e.g. colors.py?

alphastats/gui/pages/06_⚫ Results.py
```diff
@@ -61,16 +53,10 @@ def download_preprocessing_info(plot, count):
         col1, col2, col3 = st.columns([1, 1, 1])
 
         with col1:
-            save_plotly_to_pdf(plot)
-            # st.button('Save as pdf', key = "save_to_pdf" + count)
-            # if "save_to_pdf" + count in st.session_state:
-            #     save_plotly_to_pdf(plot)
+            save_plotly(plot, format="pdf")
```
straussmaximilian, line 56: Much cleaner, very nice!

alphastats/gui/utils/analysis_helper.py
```diff
@@ -26,6 +27,49 @@ def read_uploaded_file_into_df(file):
...
+            "Protein.Ids",
+            "Protein.Names",
+            "Genes",
+            "First.Protein.Description",
+            "contamination_library",
+        ]
+        if (set(expected_columns).issubset(set(df.columns.to_list()))) == False:
+            st.error("This is not a valid DIA-NN file.")
+
+    elif software == "Fragpipe":
```
straussmaximilian, line 64: Fragpipe -> FragPipe (as we also have MaxQuant and AlphaPept)

alphastats/gui/utils/analysis_helper.py
```diff
@@ -66,44 +113,88 @@ def get_analysis_options_from_dict(method, options_dict):
...
+        chosen_parameter_dict.update(
+            {"column": group, "group1": group1, "group2": group2}
+        )
 
-        group1 = st.selectbox("Group 1", options=unique_values)
+    else:
+
+        col1, col2 = st.columns(2)
+
+        with col1:
```
straussmaximilian, line 155: In case you are not aware of this you don't need to use the context `with` but can replace the `st.` with the respective `col`:

```
c1,c2 = st.columns(2)

c1.write('xyz')
```

alphastats/gui/utils/ui_helper.py
```diff
@@ -1,8 +1,34 @@
...
+# https://github.com/streamlit/streamlit/issues/4984
+
+
+def sidebar_logo():
+    img_center = """
+    <head>  
+    <title> CSS object-position property </title>  
+    <style> 
+    img {
+       position: absolute;
```
straussmaximilian, line 16: Maybe check how this looks for different displays / maybe on phone?

alphastats/test.py
```diff
@@ -0,0 +1,41 @@
+from ast import Not
```
straussmaximilian, line 1: Is this file needed?

straussmaximilian (approved):
Good changes!

PR #275, opened 2024-05-17
fix for the zeros / NA discrepancy 

boopthesnoot:
Some of the tools' output which is used in alphapeptstats use zero meaning missing data, some use NA. Previously they were not unified so the preprocessing in `_remove_na_values` was wrong. Other changes are so the downstream code can work with NAs; style, and tests.

### DIFF
--- a/alphastats/DataSet_Preprocess.py
+++ b/alphastats/DataSet_Preprocess.py
@@ -158,20 +158,20 @@ class Preprocess:
         )
         self.preprocessing_info.update({"Imputation": method})
 
-    def _linear_normalization(self, array):
+    def _linear_normalization(self, dataframe: pd.DataFrame):
         """Normalize data using l2 norm without breaking when encoutering nones
         l2 = sqrt(sum(x**2))
 
         Args:
-            array (pd.Series): array to normalize (1D array)
+            dataframe (pd.DataFrame): dataframe to normalize
 
         Returns:
-            np.array: normalized array
+            np.array: normalized np.array
         """
-        square_sum_per_row = array.pow(2).sum(axis=1, skipna=True)
+        square_sum_per_row = dataframe.pow(2).sum(axis=1, skipna=True)
 
         l2_norms = np.sqrt(square_sum_per_row)
-        normalized_vals = array.div(l2_norms.replace(0, 1), axis=0)
+        normalized_vals = dataframe.div(l2_norms.replace(0, 1), axis=0)
         return normalized_vals.values
 
     @ignore_warning(UserWarning)
--- a/alphastats/DataSet_Preprocess.py
+++ b/alphastats/DataSet_Preprocess.py
@@ -1,15 +1,15 @@
-from random import random
+import itertools
+import logging
+
+import numpy as np
 import pandas as pd
 import sklearn
-import logging
-import numpy as np
 import sklearn.ensemble
 import sklearn.impute
-from alphastats.utils import ignore_warning
-from sklearn.experimental import enable_iterative_imputer
-import itertools
-
 import streamlit as st
+
+from sklearn.experimental import enable_iterative_imputer
+from alphastats.utils import ignore_warning
 
 
 class Preprocess:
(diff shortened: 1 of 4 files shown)


alphastats/DataSet_Preprocess.py
```diff
@@ -155,26 +158,43 @@ def _imputation(self, method: str):
...
+        """Normalize data using l2 norm without breaking when encoutering nones
+        l2 = sqrt(sum(x**2))
+
+        Args:
+            array (pd.Series): array to normalize (1D array)
+
+        Returns:
+            np.array: normalized array
+        """
+        square_sum_per_row = array.pow(2).sum(axis=1, skipna=True)
```
mschwoer, line 171: does `axis=1` work if you pass a Series?

alphastats/DataSet_Preprocess.py
```diff
@@ -155,26 +158,43 @@ def _imputation(self, method: str):
...
+
+        l2_norms = np.sqrt(square_sum_per_row)
```
mschwoer, line 173: you could check how https://numpy.org/doc/stable/reference/generated/numpy.linalg.norm.html works with `NaN`s

alphastats/DataSet_Preprocess.py
```diff
@@ -123,25 +140,11 @@ def _imputation(self, method: str):
...
 
         elif method == "randomforest":
-            randomforest = sklearn.ensemble.RandomForestRegressor(
-                max_depth=10,
-                bootstrap=True,
-                max_samples=0.5,
-                n_jobs=2,
-                random_state=0,
-                verbose=0,  #  random forest takes a while print progress
+            imp = sklearn.ensemble.HistGradientBoostingRegressor(
```
mschwoer, line 143: is this the same as  `RandomForest` ? maybe introduce this as another `method` ?

alphastats/DataSet_Preprocess.py
```diff
@@ -30,45 +31,58 @@ def preprocess_print_info(self):
         print(pd.DataFrame(self.preprocessing_info.items()))
 
     def _remove_na_values(self, cut_off):
+        if (
+            self.preprocessing_info.get("Missing values were removed")
```
mschwoer, line 35: I don't know much about this `preprocessing_info`, but it seems to store information in human-readable keys.
This could lead to problems (say, I access something as `self.preprocessing_info.get("Missing values were removed.")` (did you spot the trailing dot? ;-))
I would suggest introducing a set of string constants, e.g.
`MISSING_VALUES_REMOVED = "Missing values were removed"`
somewhere and access this store exclusively through them

(not now, just for the future)

alphastats/gui/pages/02_Import Data.py
```diff
@@ -34,9 +38,14 @@
 if "loader" not in st.session_state:
     st.session_state["loader"] = None
 
+if "gene_to_prot_id" not in st.session_state:
+    st.session_state["gene_to_prot_id"] = {}
 
-def load_options():
+if "organism" not in st.session_state:
+    st.session_state["organism"] = 9606
```
mschwoer, line 45: maybe add a comment what this magic organism is? :-)

alphastats/DataSet_Preprocess.py
```diff
@@ -30,45 +31,58 @@ def preprocess_print_info(self):
...
+            )
+            return
         cut = 1 - cut_off
-        limit = self.mat.shape[0] * cut
-        
+
+        num_samples, num_proteins = self.mat.shape
+        limit = num_samples * cut
+
+        self.mat.replace(0, np.nan, inplace=True)
```
mschwoer, line 48: in principle, there's a danger here that we consider something that is actually "zero" as "missing" (should not be a problem in real life as these are floating point numbers I guess) .. still, do we have control over that upstream component that sets the zeroes initially? If so, one could try to change the zeroes to NaN there ..

mschwoer (approved):
LGTM, some probably dumb questions asked (I don't know this code base at all ;-))

alphastats/DataSet_Preprocess.py
```diff
@@ -1,14 +1,15 @@
-from random import random
-import pandas as pd
-import sklearn
+import itertools
```
mschwoer, line 1: please don't mix formatting with any functional changes as it is sometimes hard to distinguish those changes from each other .. 
I you want, we could format the whole code base together one (cf. https://github.com/MannLabs/alphabase/issues/153), should take only 15 minutes ;-)

alphastats/DataSet_Preprocess.py
boopthesnoot: Good catch, I changed the code and didn't change the docstring when I went from Series to Dataframes, fixed

alphastats/DataSet_Preprocess.py
boopthesnoot: It's also based on multiple trees, so I don't know what would be right. I could call it gradient boosting, but this could be less known for non-technical people, although it would be more correct

alphastats/gui/pages/02_Import Data.py
boopthesnoot: Done

alphastats/DataSet_Preprocess.py
boopthesnoot: Sounds good to me, when would you have the time?

alphastats/DataSet_Preprocess.py
boopthesnoot: Couldn't get it to work properly, I don't think it works well with `NaN`s

PR #280, opened 2024-05-24
FIX: macos release crashing because of miniconda installation

boopthesnoot:
Update release_on_click_installer.yml


boopthesnoot:
the underlying issue I'm trying to fix: https://github.com/MannLabs/alphapeptstats/actions/runs/9194179735/job/25286989541

PR #291, opened 2024-08-28
Bookkeeping of session state variables

JuliaS92:
As by a suggestion from Magnus, we should introduce a class to keep track of all session_state keys used throughout the app and replace all instances of accessing the keys with a reference to this class.

Turn this:
`if 'dataset' in st.session_state`

into this:
```
class SessionStateKeys:
    DATASET = 'dataset

if SessionStateKeys.DATASET in st.session_state
```

This way we can easily look up all references to that state and we have an overview of all keys used throughout the app.

I implemented this for the data import page.

Please check if this looks as intended Magnus, and if you think this might be breaking at other points in the app Mikhail.


alphastats/gui/pages/02_Import Data.py
```diff
@@ -362,12 +362,12 @@ def empty_session_state():
...
         del st.session_state[key]
     st.empty()
-    st.session_state["software"] = "<select>"
+    st.session_state[SessionStateKeys.SOFTWARE] = "<select>"
 
     from streamlit.runtime.scriptrunner.script_run_context import get_script_run_ctx
 
     user_session_id = get_script_run_ctx().session_id
-    st.session_state["user_session_id"] = user_session_id
+    st.session_state[SessionStateKeys.USER_SESSION_ID] = user_session_id
```
mschwoer, line 370: mb rename to `StateKeys` to save some visual clutter :-)

alphastats/gui/pages/02_Import Data.py
```diff
@@ -195,21 +195,21 @@ def upload_softwarefile(software):
...
             )
-            st.session_state["loader"] = loader
+            st.session_state[SessionStateKeys.LOADER] = loader
 
 
 def create_metadata_file():
-    dataset = DataSet(loader=st.session_state.loader)
-    st.session_state["metadata_columns"] = ["sample"]
+    dataset = DataSet(loader=st.session_state[SessionStateKeys.LOADER])
+    st.session_state[[SessionStateKeys.METADATA_COLUMNS]] = ["sample"]
```
mschwoer, line 212: one pair of `[]` too much?

alphastats/gui/pages/02_Import Data.py
```diff
@@ -166,8 +166,8 @@ def select_sample_column_metadata(df, software):
         submitted = st.form_submit_button("Create DataSet")
 
     if submitted:
-        if len(df[st.session_state.sample_column].to_list()) != len(
-            df[st.session_state.sample_column].unique()
+        if len(df[st.session_state[SessionStateKeys.SAMPLE_COLUMN]].to_list()) != len(
```
mschwoer, line 169: +1 for unifying the access (attribute vs. element)

alphastats/gui/pages/02_Import Data.py
```diff
@@ -29,26 +29,26 @@
...
-    st.session_state["organism"] = 9606  # human
+    st.session_state[SessionStateKeys.ORGANISM] = 9606  # human
 
 
 def load_options():
     from alphastats.gui.utils.options import plotting_options, statistic_options
 
-    st.session_state["plotting_options"] = plotting_options
-    st.session_state["statistic_options"] = statistic_options
+    st.session_state[SessionStateKeys.PLOTTING_OPTIONS] = plotting_options
```
mschwoer, line 50: high level view: there is _a lot_ happening with the session state already only on this page .. maybe we should also revisit its use and try to limit it as much as possible

JuliaS92: Once I am done with the bookkeeping and the page tests that's exactly what I plan to do, also to make the saving and loading of settings more straighforward.

mschwoer:
done in https://github.com/MannLabs/alphapeptstats/pull/326

PR #292, opened 2024-08-29
Draft PR to get feedback on first tests

JuliaS92:
I have used the streamlit testing library in combination with mocking of file inputs (thank you @mschwoer for help with that).

I now have 4 tests so far:
- page loads
- page loads with mocking
- sample dataset loads
- maxquant testdata (proteinGroups and metadata) loads

In each case I have only asserted the session state and not what the interface actually displays.

This way, we can test if the app runs through and can also make sure that what's persistent between pages is correct, while not affecting testing by aestetic changes to the interface.

I would further be happy to receive suggestions on how to best structure the tests to not duplicate too much code, if I add a test for each software option. (A link to a good example for shared/parameterized setup would be appreciated.)

I will also add tests to make sure the procedurally generated widgets are populated correctly, unless you think this is unnecessary work.

I would not go into testing different input combinations too much, as this is covered by the loader unittests.


alphastats/gui/test/test_02_Import Data.py
```diff
@@ -0,0 +1,100 @@
...
+    at.run()
+    assert at.session_state.dataset.gene_names == "Gene names"
+    assert at.session_state.dataset.index_column == "Protein IDs"
+    assert at.session_state.dataset.intensity_column == 'LFQ intensity [sample]'
+    assert at.session_state.dataset.rawmat.shape == (312, 2611)
+    assert at.session_state.dataset.software == 'MaxQuant'
+    assert at.session_state.dataset.sample == 'sample'
+    assert str(type(at.session_state.distribution_plot)) == "<class 'plotly.graph_objs._figure.Figure'>"
+    assert str(type(at.session_state.loader)) == "<class 'alphastats.loader.MaxQuantLoader.MaxQuantLoader'>"
+    assert "plotting_options" in at.session_state
```
mschwoer, line 95: just FYI: if possible, aim for testing the specific values (`assert at.session_state["plotting_options"] == {...}`, rather than just checking if the key exists.. makes the test stronger

mschwoer (commented):
🚀

alphastats/gui/AlphaPeptStats.py
```diff
@@ -17,7 +17,7 @@
 
 runtime = get_instance()
 session_id = get_script_run_ctx().session_id
-session_info = runtime._session_mgr.get_session_info(session_id)
+# session_info = runtime._session_mgr.get_session_info(session_id)
```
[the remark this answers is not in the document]
JuliaS92: Yes, this line causes the testing framework to fail and so far I haven't found the purpose otherwise. If we really don't need the line we can delete it at some point.

alphastats/gui/test/test_02_Import Data.py
```diff
@@ -0,0 +1,100 @@
...
+    at.run()
+
+    assert at.session_state.organism == 9606
+    assert at.session_state.user_session_id == 'test session id'
+    assert at.session_state.software == '<select>'
+    assert at.session_state.gene_to_prot_id == {}
+    assert at.session_state.loader == None
+
+@patch("streamlit.file_uploader")
+def test_loadmockedpage(mock_file: MagicMock):
```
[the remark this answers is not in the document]
JuliaS92: The only difference is that this time the file_uploader is mocked. This is to make sure the MaigMock doesn't cause problems in itself.

alphastats/gui/test/test_02_Import Data.py
```diff
@@ -0,0 +1,100 @@
...
+    assert at.session_state.user_session_id == 'test session id'
+    assert at.session_state.software == '<select>'
+    assert at.session_state.gene_to_prot_id == {}
+    assert at.session_state.loader == None
+
+def test_sampledata():
+    at = AppTest(APP_FILE, default_timeout=200)
+    at.run()
+
+    at.button[0].click().run()
```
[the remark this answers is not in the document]
JuliaS92: For most elements there is, but some are procedurally generated and contain a unique string - like the buttons in forms. I will add a comment.

alphastats/gui/utils/analysis_helper.py
```diff
@@ -134,7 +134,7 @@ def st_general(method_dict):
 
 @st.cache_data
 def gui_volcano_plot_differential_expression_analysis(
-    chosen_parameter_dict, user_session_id=st.session_state.user_session_id
+    chosen_parameter_dict, user_session_id
 ):
     """
     initalize volcano plot object with differential expression analysis results
```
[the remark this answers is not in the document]
JuliaS92: So the number of fields displayed is always the same and it is just a matter of populating the options after the file is uploaded. I will change this the behaviour to show the whole interface (both file_loaders and all three dropdowns) at once and then just update the options when the files are loaded.

alphastats/gui/test/test_02_Import Data.py
JuliaS92: This was my way of vertical slicing ;) The values of these operations are actually tested elsewhere, although I don't have a full overview yet. For now this is just to make sure the function returned the correct object and not None or an error.

alphastats/gui/utils/analysis_helper.py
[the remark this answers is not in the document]
mschwoer: if this looked like this (pseudocode)
```
value1, value2, value3 = get_values_for_populating_options(..., selected_software="something")

st.field1.value = value1
st.field2.value = value2
...
```
then 1 UI test with a mocked `get_values_for_populating_options` that checks that the fields are correctly populated would suffice ..
`get_values_for_populating_options` could then be tested outside of the streamlit testing framework

alphastats/gui/AlphaPeptStats.py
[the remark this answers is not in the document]
mschwoer: if we don't use `session_info` anywhere, you can also just delete it ;-)

alphastats/gui/pages/02_Import Data.py
```diff
@@ -207,6 +210,8 @@ def upload_softwarefile(software):
             st.session_state["loader"] = loader
 
 
+# TODO: Move this to new data_helper.py in utils
+# TODO: Make this more powerful by making suggestions based on the sample names
```
mschwoer, line 214: The "more powerful" TODO I would move to a ticket

alphastats/gui/test/test_02_import_data.py
```diff
@@ -0,0 +1,135 @@
...
+import pandas as pd
+from io import BytesIO
+
+def print_session_state(apptest: AppTest):
+    for k,v in apptest.session_state.filtered_state.items():
+        print(f"{k}:    {str(type(v))}   {str(v)[:20] if type(v) not in [int, list, str] else v}")
+
+APP_FOLDER = Path(__file__).parent / Path("../")
+APP_FILE = f"{APP_FOLDER}/pages/02_Import Data.py"
+TEST_INPUT_FILES = f"{APP_FOLDER}/../../testfiles"
```
mschwoer, line 13: `TEST_INPUT_FILES_PATH = APP_FOLDER / "../../testfiles"

alphastats/gui/test/test_02_import_data.py
```diff
@@ -0,0 +1,135 @@
...
+from unittest.mock import MagicMock, patch
+import pandas as pd
+from io import BytesIO
+
+def print_session_state(apptest: AppTest):
+    for k,v in apptest.session_state.filtered_state.items():
+        print(f"{k}:    {str(type(v))}   {str(v)[:20] if type(v) not in [int, list, str] else v}")
+
+APP_FOLDER = Path(__file__).parent / Path("../")
+APP_FILE = f"{APP_FOLDER}/pages/02_Import Data.py"
```
mschwoer, line 12: could also use Path concatenation with `/` here

alphastats/gui/test/test_02_import_data.py
```diff
@@ -0,0 +1,135 @@
...
+    assert "statistic_options" in at.session_state
+
+
+def _data_buf(path_from_testfiles: str):
+    """Helper function to open a data file from the testfiles folder and return a BytesIO object.
+    
+    Additionally add filename as attribute."""
+    with open(f"{TEST_INPUT_FILES}{path_from_testfiles}", "rb") as f:
+        buf = BytesIO(f.read())
+        buf.name = path_from_testfiles.split('/')[-1]
```
mschwoer, line 67: could use pathlib here: `.stem` or `.name` or smth

(and in the `with` statement)

alphastats/gui/test/test_02_import_data.py
```diff
@@ -0,0 +1,135 @@
...
+    assert at.session_state.metadata_columns == ['sample', 'disease', 'Drug therapy (procedure) (416608005)', 'Lipid-lowering therapy (134350008)']
+    assert str(type(at.session_state.dataset)) == "<class 'alphastats.DataSet.DataSet'>"
+    assert at.session_state.software == "<select>"
+    assert str(type(at.session_state.distribution_plot)) == "<class 'plotly.graph_objs._figure.Figure'>"
+    assert str(type(at.session_state.loader)) == "<class 'alphastats.loader.MaxQuantLoader.MaxQuantLoader'>"
+    assert "plotting_options" in at.session_state
+    assert "statistic_options" in at.session_state
+
+
+def _data_buf(path_from_testfiles: str):
```
mschwoer, line 61: (nit) return value type hint

alphastats/gui/test/test_02_import_data.py
```diff
@@ -0,0 +1,135 @@
+from streamlit.testing.v1 import AppTest
+from pathlib import Path
+from unittest.mock import MagicMock, patch
+import pandas as pd
+from io import BytesIO
+
+def print_session_state(apptest: AppTest):
```
mschwoer, line 7: is this used somewhere? also, redefined in `test_03...`

mschwoer (approved):
LGTM! some comments around the consistent use of `pathlib.Path`

straussmaximilian (approved):
Thanks for taggin me. 
App-Testing in streamlit is very nice and I was not aware of this :)

PR #306, opened 2024-09-06
refactoring_import

mschwoer:
Quite a massive change, tried to limit it to page 02
- Refactor data import to contain only display logic, simplify structure 
- streamline inport user flow
- Move display of state info to page 03
- Reduce use of session state
- Introduce init_session_state
Note: will most likely fail some tests created here: https://github.com/MannLabs/alphapeptstats/pull/305


alphastats/gui/pages/02_Import Data.py
```diff
@@ -1,14 +1,21 @@
 import streamlit as st
```
mschwoer, line 1: probably best to not look at the diff for this page, but just at the new version

alphastats/gui/pages/02_Import Data.py
```diff
@@ -1,76 +1,141 @@
...
-st.session_state["user_session_id"] = user_session_id
+st.markdown("### Start a new session")
+st.write(
+    "Start a new session will discard the current one (including all analysis!) and enable importing a new dataset."
+)
+st.write("To explore AlphaPeptStats you may also load an example dataset.")
 
-if "loader" not in st.session_state:
-    st.session_state["loader"] = None
+c1, c2 = st.columns(2)
```
JuliaS92, line 30: I like the new session or new session with example dataset.
Longterm I would however prefer that the session doesn't need to be reset 'manually', i.e. just have the interface for loading data and button to load the sample data, with no need to reset the session (this should just happen automatically)

alphastats/gui/pages/02_Import Data.py
```diff
@@ -1,76 +1,141 @@
...
+    init_session_state()
+    loader, metadata_columns, dataset = load_example_data()
 
-if "organism" not in st.session_state:
-    st.session_state["organism"] = 9606  # human
+    st.session_state["dataset"] = dataset
+    st.session_state["metadata_columns"] = metadata_columns
+    st.session_state["loader"] = loader
+    load_options()
+    # TODO why are we doing this so early?
```
JuliaS92, line 44: This was just to show a plot of the data on upload. I think a button to switch to Data Overview page would be preferable.

alphastats/gui/pages/02_Import Data.py
```diff
@@ -1,76 +1,141 @@
...
+    st.stop()
 
-sidebar_info()
 
-if "dataset" not in st.session_state:
-    st.markdown("### Import Proteomics Data")
+st.markdown("### Import Proteomics Data")
+if "dataset" in st.session_state:
+    st.info(f"DataSet already present: {st.session_state['dataset']}")
+    st.stop()
```
JuliaS92, line 53: I wasn't aware this exists and we should use it more often.

alphastats/gui/pages/02_Import Data.py
```diff
@@ -1,76 +1,141 @@
...
+    intensity_column=intensity_column,
+    index_column=index_column,
+    software=software,
+)
+
+
+# ##########  Load Metadata File
+st.markdown("##### 3. Prepare Metadata (optional)")
+sample_column = None
+metadatafile_df = show_metadata_file_uploader(loader)
```
JuliaS92, line 101: This should be consistent with the software file uploader. Either have both fileuploaders in the page file (preferred) or both in the helper file.

alphastats/gui/pages/03_Data Overview.py
```diff
@@ -1,6 +1,8 @@
 import streamlit as st
-import pandas as pd
-import plotly.express as px
+
+from alphastats.gui.utils.import_helper import (
+    display_loaded_dataset,
```
JuliaS92, line 4: move this function to ui helper, to make it clear that it is used by more than one page.

alphastats/gui/utils/analysis_helper.py
```diff
@@ -79,26 +85,23 @@ def download_preprocessing_info(plot):
     )
 
 
-def read_uploaded_file_into_df(file, decimal="."):
-    filename = file.name
+def _read_file_to_df(file: UploadedFile, decimal: str = ".") -> Optional[pd.DataFrame]:
```
JuliaS92, line 88: Move this to ui_helper, too if it is used by several pages

alphastats/gui/utils/analysis_helper.py
```diff
@@ -79,26 +85,23 @@ def download_preprocessing_info(plot):
...
-        logging.warning(
-            "WARNING: File could not be read. \nFile has to be a .xslx, .tsv, .csv or .txt file"
-        )
-        return
+    elif extension == ".csv":
+        return pd.read_csv(file, decimal=decimal)
 
-    return df
+    raise ValueError(
+        f"Unknown file type '{extension}'. \nSupported types: .xslx, .tsv, .csv or .txt file"
```
JuliaS92, line 103: How does this appear in the logging output in the end? Do we even need the logging library?

alphastats/gui/utils/analysis_helper.py
```diff
@@ -364,24 +367,22 @@ def helper_compare_two_groups():
     return chosen_parameter_dict
 
 
-def get_sample_names_from_software_file():
+def get_sample_names_from_software_file(loader: BaseLoader) -> List[str]:
```
JuliaS92, line 370: why is this function in the anlaysis helper?

alphastats/gui/utils/options.py
```diff
@@ -150,7 +150,7 @@ def statistic_options(state):
         "import_file": "Please upload results_proteins.csv or results.hdf",
         "intensity_column": ["[sample]_LFQ"],
         "index_column": ["Unnamed: 0"],
-        "loader_function": AlphaPeptLoader,
+        "loader_function": AlphaPeptLoader,  # TODO loader_function -> loader_class
```
JuliaS92, line 153: maybe just loader for future compatibility when we might use something else?

tests/gui/test_02_import_data.py
```diff
@@ -52,7 +47,7 @@ def test_page_02_loads_sample_data():
     at.run()
 
     # User clicks Load Sample Data button
-    at.button("load_sample_data").click().run()
+    at.button[1].click().run()
```
JuliaS92, line 50: Please give a key to the button instead so the test works if the button order changes.

tests/gui/test_02_import_data.py
```diff
@@ -112,7 +105,7 @@ def test_page_02_loads_maxquant_testfiles(mock_file_uploader: MagicMock):
     at.run()
 
     # User selects MaxQuant from the dropdown menu
-    at.selectbox(key="software").select("MaxQuant")
+    at.selectbox[0].select("MaxQuant")
```
JuliaS92, line 108: I know the goal was to remove unnecessary entries in the session state. However I see two reasons for using the key, i.e. keeping them in the session state: 1. we can address them in tests regardless of itnerface rearrangements, 2. we can set their value on loading settings. They will be part of the session_state with or without a key, but only with a key they can be addressed easily.

tests/gui/test_02_import_data.py
```diff
@@ -129,33 +122,23 @@ def test_page_02_loads_maxquant_testfiles(mock_file_uploader: MagicMock):
 
     assert not at.exception
 
-    assert (
```
JuliaS92, line 132: I think it would eb important to test for each software, that the widgets are correctly initialized.

alphastats/gui/pages/02_Import Data.py
```diff
@@ -1,76 +1,141 @@
...
-if st.button("New Session: Import new dataset"):
-    empty_session_state()
-    st.rerun()
+softwarefile_df = load_softwarefile_df(software, softwarefile)
+
+intensity_column, index_column = show_loader_columns_selection(
+    software=software, softwarefile_df=softwarefile_df
+)
+
+loader = load_proteomics_data(
```
JuliaS92, line 90: It could be that we actually need some logic here for the custom upload, which serves the multiselect instead of selector. Intensity columns would then be None initially.

alphastats/gui/pages/02_Import Data.py
```diff
@@ -1,76 +1,141 @@
...
+    sidebar_info()
+    st.stop()
 
-sidebar_info()
 
-if "dataset" not in st.session_state:
-    st.markdown("### Import Proteomics Data")
+st.markdown("### Import Proteomics Data")
+if "dataset" in st.session_state:
+    st.info(f"DataSet already present: {st.session_state['dataset']}")
```
JuliaS92, line 52: It would be great to provide additional info here and maybe show the preview. Also maybe refer to the sidebar and the data overview.

alphastats/gui/pages/02_Import Data.py
```diff
@@ -1,76 +1,141 @@
...
+    dataset = DataSet(
+        loader=loader,
+        metadata_path=metadatafile_df,
+        sample_column=sample_column,
+    )
+    metadata_columns = metadatafile_df.columns.to_list()
+
+if dataset is not None:
+    st.info("DataSet has been created.")
+    st.session_state["dataset"] = dataset
```
JuliaS92, line 134: Maybe bundle this into a function?

alphastats/gui/pages/02_Import Data.py
```diff
@@ -1,76 +1,141 @@
...
+    st.session_state["metadata_columns"] = metadata_columns
+    st.session_state["loader"] = loader
```
JuliaS92, line 136: TODO: Figure out if we even need the loader here, as the dataset has the loader as an attribute.

alphastats/gui/utils/import_helper.py
```diff
@@ -233,143 +188,141 @@ def empty_session_state():
...
+    ]
 
     if len(valid_sample_columns) == 0:
-        st.error(
+        raise ValueError(
             f"Metadata does not match Proteomics data."
             f"Information for the samples: {samples_proteomics_data} is required."
         )
 
+    # TODO I get an ERROR: "described in Please upload proteinGroups.txt"
```
JuliaS92, line 305: This is unrelated to the software and can just say 'described in the column names of the uploaded proteomic data file.

JuliaS92 (approved):
LGTM - I like how everything is now a lot more modular and the flow can directly be deduced from the main page file :)
Some design decisions to be made though:
- I would propose keeping the keys of the widgets to make them uniquely accessible by key
- I would propose moving any helper functions shared by more than one page to ui_helper so we always know of the other functions that they only break the one page and not multiple ones.

alphastats/gui/pages/02_Import Data.py
mschwoer: done in https://github.com/MannLabs/alphapeptstats/pull/307

alphastats/gui/utils/analysis_helper.py
mschwoer: done in https://github.com/MannLabs/alphapeptstats/pull/307

alphastats/gui/pages/03_Data Overview.py
mschwoer: done in https://github.com/MannLabs/alphapeptstats/pull/307

alphastats/gui/utils/analysis_helper.py
mschwoer: the error message should be logged automatically .. the logging lib still has value to add some `INFO` statements that might help

alphastats/gui/utils/analysis_helper.py
mschwoer: moved in https://github.com/MannLabs/alphapeptstats/pull/307

alphastats/gui/pages/02_Import Data.py
mschwoer: removed in https://github.com/MannLabs/alphapeptstats/pull/307

alphastats/gui/pages/02_Import Data.py
mschwoer: I would not double UI elements for now, will add a link to the "data overview" page

alphastats/gui/pages/02_Import Data.py
mschwoer: done in #307

tests/gui/test_02_import_data.py
mschwoer: yup, but e.g. the `loader` is not in the session stage at this stage

tests/gui/test_02_import_data.py
mschwoer: agree.. done in #307
(added a trailing underscore to tell these widget keys apart from other session keys .. will get unified during the bookkeeping tasks)

PR #307, opened 2024-09-09
Refactoring import ii

mschwoer:
following https://github.com/MannLabs/alphapeptstats/pull/306/files#r1749959452:
get rid if saving one plot in the session state upfront


alphastats/gui/utils/overview_helper.py
```diff
@@ -12,13 +13,18 @@ def get_sample_histogram_matrix(user_session_id):
...
+
+
 @st.cache_data
 def get_intensity_distribution_processed(user_session_id):
     return st.session_state.dataset.plot_sampledistribution()
 
 
 @st.cache_data
-def get_display_matrix(user_session_id):
+def get_display_matrix():
```
JuliaS92, line 27: we need at least one argument for every cached funciton or the output will never change.

alphastats/DataSet_Plot.py
```diff
@@ -199,21 +199,27 @@ def plot_correlation_matrix(self, method: str = "pearson"):
         return plot
 
     def plot_sampledistribution(
-        self, method: str = "violin", color: str = None, log_scale: bool = False
+        self,
+        method: str = "violin",
+        color: str = None,
+        log_scale: bool = False,
+        use_raw: bool = False,
```
JuliaS92, line 206: That's very helpful

JuliaS92 (approved):
LGTM

alphastats/gui/utils/overview_helper.py
mschwoer: ah, thanks, got it

alphastats/gui/pages/02_Import Data.py
```diff
@@ -61,8 +79,7 @@
 options = [default_select_option] + list(SOFTWARE_OPTIONS.keys())
 
 software = st.selectbox(
-    "Select your Proteomics Software",
-    options=options,
+    "Select your Proteomics Software", options=options, key="_software"
```
JuliaS92, line 82: Do I get it right, that you want to use underscores to mark widget keys, as opposed to manually set session_states like dataset?

JuliaS92 (approved):
Looks great :)

PR #308, opened 2024-09-09
Live representation of selected steps in predefined workflow

JuliaS92:
I added a cytoscape graph to represent the current processing workflow.
To make updatable without hitting the submit button, I removed the forms on the processing page.
In the process I integrated the batch correction into the preprocessing pipeline. To enable rerunning from half the pipeline one could add a button to rerun from changed settings in the future (e.g. different batch correction, while not rerunning random forest imputation).

I would also add the reverse update to make clicking a node change its state.


alphastats/gui/pages/03_Preprocessing.py
```diff
@@ -1,16 +1,88 @@
...
+        (
+            remove_contaminations,
+            remove_samples,
+            subset,
+            data_completeness,
+            log2_transform,
+            normalization,
+            imputation,
+            batch,
+        ) = configure_preprocessing()
```
mschwoer, line 45: what if we introduce a dictionary that maps all preprocessing steps to their parameters?

```python
mapping = {
"imputation": "median",
"log2_transform": False,
"normalization": None
...
}
```
this has several benefits:
- only one argument is passed around
- order of imputations can be encoded
- serialization is easier

alphastats/gui/utils/preprocessing_helper.py
```diff
@@ -1,141 +1,255 @@
...
 import datetime
 
 
-def preprocessing():
-    c1, c2 = st.columns(2)
+def draw_predefined_workflow(
+    workflow=["remove contaminations", "subset data", "log2 transform"],
+):
+    available_steps = [
+        "remove contaminations",
```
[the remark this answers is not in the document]
JuliaS92: I was thinking of introducing a workflow class, which would contain:
- constants for available steps (maybe a dictionary mapping a step identifier to the ui representation, help text and backend method)
- an ordered list of selected steps
- a dictionary with additional settings required for each step

The preprocessing function would then turn into something that just takes a Workflow instance and maybe a continue flag as input.

If that's fine by you I would defer this to a separate PR.

alphastats/gui/utils/preprocessing_helper.py
```diff
@@ -1,141 +1,255 @@
...
 import pandas as pd
+from st_cytoscape import cytoscape
 
 import datetime
 
 
-def preprocessing():
-    c1, c2 = st.columns(2)
+def draw_predefined_workflow(
+    workflow=["remove contaminations", "subset data", "log2 transform"],
```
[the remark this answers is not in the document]
JuliaS92: No we actually don't, this was just a remnant of a previous commit.

alphastats/gui/utils/preprocessing_helper.py
```diff
@@ -1,141 +1,255 @@
...
+                remove_samples,
+                subset,
+                data_completeness,
+                log2_transform,
+                normalization,
+                imputation,
+                batch,
+            ],
+        )
+        if form not in [None, False, [], 0.0]
```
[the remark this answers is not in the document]
JuliaS92: are you fine with if bool(form) to make it clear that we are going by thruthiness and not specifically True?

alphastats/gui/pages/03_Preprocessing.py
JuliaS92: I would go straight for returning a workflow object, once this is designed (https://github.com/MannLabs/alphapeptstats/pull/308#discussion_r1751477966)

alphastats/gui/utils/preprocessing_helper.py
```diff
@@ -1,141 +1,255 @@
...
+        st.session_state.dataset.batch_correction(batch=batch)
 
 
-def main_preprocessing():
-    if "dataset" in st.session_state:
-        preprocessing()
-
-    else:
-        st.info("Import Data first")
+def reset_preprocessing():
```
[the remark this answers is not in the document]
JuliaS92: I opted for extracting the button to the page file.

alphastats/gui/utils/preprocessing_helper.py
```diff
@@ -1,141 +1,255 @@
...
+
+    subset = st.selectbox(
+        "Subset data so it matches with metadata. Remove miscellanous samples in rawinput.",
+        options=[True, False],
+    )
+
+    remove_samples = st.multiselect(
+        "Remove samples from analysis",
+        options=[]
+        if "dataset" not in st.session_state
```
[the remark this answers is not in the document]
JuliaS92: My idea was, that one can explore the app and it's options without being forced to load a dataset at all (essentially removing the need to load the sample data).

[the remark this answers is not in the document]
JuliaS92: How about passing dataset as an argument?

alphastats/gui/utils/preprocessing_helper.py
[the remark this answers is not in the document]
JuliaS92: https://github.com/MannLabs/alphapeptstats/pull/308/commits/460d3d03394667426b7a455814a6b859a9b8de64

Two new constants for all the strings.

alphastats/gui/pages/03_Preprocessing.py
JuliaS92: https://github.com/MannLabs/alphapeptstats/pull/308/commits/4dad21d8017ebb8e4574f1b564d71c7dba1c3ad6

alphastats/gui/utils/preprocessing_helper.py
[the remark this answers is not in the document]
JuliaS92: I added dataset as argument, but only call it if data was actually loaded, so no logic is needed

alphastats/gui/pages/03_Preprocessing.py
JuliaS92: https://github.com/MannLabs/alphapeptstats/pull/308/commits/f477837bf9467165232ff7db4c17e7880664f92b

now also using the dictionary

alphastats/gui/utils/preprocessing_helper.py
```diff
@@ -1,124 +1,225 @@
...
+        "help": 'Normalize data using one of the available methods ("zscore", "quantile", "vst", "linear").',
+    },
+    "imputation": {
+        "repr": "Imputation",
+        "help": 'Impute missing values using one of the available methods ("mean", "median", "knn", "randomforest").',
+    },
+    "batch": {"repr": "Batch correction", "help": "Batch correction."},
+}
+
+PREDEFINED_ORDER = [
```
mschwoer, line 76: could just use the order from `WORKFLOW_STEPS.keys()` instead

alphastats/gui/utils/preprocessing_helper.py
```diff
@@ -1,124 +1,225 @@
...
+    return new_workflow
+
+
+def run_preprocessing(
+    settings,
+):
+    settings["remove_samples"] = (
+        settings["remove_samples"] if len(settings["remove_samples"]) != 0 else None
+    )
+    st.session_state.dataset.preprocess(**settings)
```
mschwoer, line 206: could you add a TODO here to pass the whole `settings` object and introduce string constants to access it? (this way it's a bit fragile as the parameters of the `preprocess` method need to be compatible with some strings defined completely elsewhere (`configure_preprocessing()`)

alphastats/gui/utils/preprocessing_helper.py
```diff
@@ -1,124 +1,225 @@
...
+    new_workflow = [
+        key
+        for key, setting in settings.items()
+        if bool(setting)  # use of falsiness
+    ]
+    return new_workflow
+
+
+def run_preprocessing(
+    settings,
```
mschwoer, line 201: pass `dataset` as parameter?

alphastats/gui/utils/preprocessing_helper.py
```diff
@@ -1,124 +1,225 @@
...
+):
+    settings["remove_samples"] = (
+        settings["remove_samples"] if len(settings["remove_samples"]) != 0 else None
+    )
+    st.session_state.dataset.preprocess(**settings)
+
+    preprocessing = st.session_state.dataset.preprocessing_info
```
mschwoer, line 208: please add a TODO to have `.preprocess(**settings)` return this object

alphastats/gui/utils/preprocessing_helper.py
```diff
@@ -1,124 +1,225 @@
...
+    "subset",
+    "data_completeness",
+    "log2_transform",
+    "normalization",
+    "imputation",
+    "batch",
+]
+
+
+def draw_workflow(workflow: list[str], order: list[str] = PREDEFINED_ORDER):
```
mschwoer, line 88: could move this and `CYTOSCAPE_STYLESHEET` to a new module `preprocessing_drawer` or smth

mschwoer (approved):
LGRM

alphastats/gui/utils/preprocessing_helper.py
JuliaS92: I agree on the stylesheet, but not the function. I would leave it as is now unless it really bothers you.

alphastats/gui/utils/preprocessing_helper.py
JuliaS92: That would be breaking for the sidebar, as this is displayed there. So we need this available on all pages.

alphastats/gui/utils/preprocessing_helper.py
JuliaS92: https://github.com/MannLabs/alphapeptstats/pull/308/commits/ca3e69ffcc24f316faa5220c94c414f528446a07

alphastats/gui/utils/preprocessing_helper.py
JuliaS92: I'd rather keep this separate, so new workflow steps can be added without regarding the order.

alphastats/gui/pages/03_Preprocessing.py
```diff
@@ -1,16 +1,65 @@
...
+        PREPROCESSING_STEPS.SUBSET["key"],
+        PREPROCESSING_STEPS.LOG2_TRANSFORM["key"],
+    ]
+
 st.markdown("### Preprocessing")
+c1, c2 = st.columns([1, 1])
+
+with c2:
+    if "dataset" not in st.session_state:
+        settings = {k: True for k in st.session_state.workflow}
```
mschwoer, line 31: seems a bit hacky .. 
the explicit way would be
```
def update_workflow(
    settings: dict, show_all: bool = False
):
    new_workflow = [
        key
        for key, setting in settings.items()
        if show_all or setting
    ]
    return new_workflow
```

alphastats/gui/utils/preprocessing_helper.py
```diff
@@ -1,144 +1,255 @@
...
+    PREPROCESSING_STEPS.DATA_COMPLETENESS["key"],
+    PREPROCESSING_STEPS.LOG2_TRANSFORM["key"],
+    PREPROCESSING_STEPS.NORMALIZATION["key"],
+    PREPROCESSING_STEPS.IMPUTATION["key"],
+    PREPROCESSING_STEPS.BATCH["key"],
+]
+
+
+def draw_workflow(workflow: list[str], order: list[str] = PREDEFINED_ORDER):
+    def find_step_by_key_value(key, target_value):
```
mschwoer, line 104: hm, why did we make `PREPROCESSING_STEPS` a class? could as well be a `dict` of `dicts` ..

JuliaS92: https://github.com/MannLabs/alphapeptstats/pull/308/commits/112a6d98626d23b84a3371c5077a0e449365a97d

alphastats/gui/pages/03_Preprocessing.py
JuliaS92: https://github.com/MannLabs/alphapeptstats/pull/308/commits/e398d76ada357c5e40276a1e0b1bea6432b1b170

It was entirely unnecessary to even create settings if no dataset is present.

PR #309, opened 2024-09-09
Bump streamlit

mschwoer:
Bump streamlit to latest.
- Enable circumvention of `AxiosError`
- Determine user session id ourselves
- Make all variables required for caching explicit

### DIFF
--- a/alphastats/gui/pages/02_Import Data.py
+++ b/alphastats/gui/pages/02_Import Data.py
@@ -7,16 +7,18 @@ from alphastats.gui.utils.options import SOFTWARE_OPTIONS
 
 from alphastats.gui.utils.import_helper import (
     load_example_data,
-    empty_session_state,
     uploaded_file_to_df,
     show_loader_columns_selection,
     load_proteomics_data,
     load_options,
     show_select_sample_column_for_metadata,
-    init_session_state,
     show_button_download_metadata_template_file,
 )
-from alphastats.gui.utils.ui_helper import sidebar_info
+from alphastats.gui.utils.ui_helper import (
+    sidebar_info,
+    empty_session_state,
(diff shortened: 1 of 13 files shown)


alphastats/gui/pages/02_Import Data.py
```diff
@@ -57,14 +59,14 @@ def _finalize_data_loading(
     loader, metadata_columns, dataset = load_example_data()
 
     _finalize_data_loading(loader, metadata_columns, dataset)
-    # st.page_link("pages/03_Data Overview.py", label="=> Go to data overview..")  # TODO: needs newer streamlit
+    st.page_link("pages/03_Data Overview.py", label="=> Go to data overview..")
```
JuliaS92, line 62: maybe make this a part of _finalize_data_loading?

tests/test_DataSet.py
```diff
@@ -520,8 +520,7 @@ def test_plot_intenstity_subgroup(self):
         plot_dict = plot.to_plotly_json()
         self.assertEqual(len(plot_dict.get("data")), 3)
 
-    @patch("logging.Logger.warning")
-    def test_plot_intenstity_subgroup_significance_warning(self, mock):
+    def test_plot_intensity_subgroup_gracefully_handle_one_group(self):
```
JuliaS92, line 523: For understanding: we just let it raise the warning instead of mocking it?

JuliaS92: Should we then assert that the specific warning this tests for is raised, since this was the original intention of self.assertEqual(mock.call_count, 1)?

mschwoer: nah, not worth the effort

JuliaS92 (approved):

PR #310, opened 2024-09-11
Testpage3

JuliaS92:
These are simple tests to run the preprocessing page, without checking the actual result.

I assume that all preprocessing functions are tested elsewhere.

In the next PR one should add tests to detect desynchronization of the interface and session_state.


mschwoer (approved):
LGTM

PR #314, opened 2024-09-12
disable caching

mschwoer:
disable caching to quick-fix the related bugs


JuliaS92 (approved):
Added comment: This also moved covert_df to the ui_helper and removed it in all other files. Functions that were previously cached also don't have the user_session_id as argument anymore.

PR #319, opened 2024-09-12
Tests page 3

JuliaS92:
Functional test for page 3


tests/gui/conftest.py
```diff
@@ -0,0 +1,27 @@
...
+from alphastats.load_data import load_data
+from pathlib import Path
+from alphastats import DataSet
+
+
+APP_FOLDER = Path(__file__).parent / Path("../../alphastats/gui/")
+TEST_INPUT_FILES = f"{APP_FOLDER}/../../testfiles"
+
+
+def print_session_state(apptest: AppTest):
```
mschwoer, line 11: good idea collecting these also here!

but there are some redefinitions now that should be subsituted by these
e.g. in test_03_data_overview.py

mschwoer (approved):

tests/gui/conftest.py
```diff
@@ -0,0 +1,27 @@
...
+    for k, v in apptest.session_state.filtered_state.items():
+        print(
+            f"{k}:    {str(type(v))}   {str(v)[:20] if type(v) not in [int, list, str] else v}"
+        )
+
+
+def create_dataset_alphapept():
```
mschwoer, line 18: by convention, conftest is at the root of the tests directory

JuliaS92: intended for helpers that are only used part of the tests

mschwoer (approved):

PR #322, opened 2024-09-12
Draft refactor dataset

mschwoer:
Starts decoupling DataSet from its superclasses.
Also, simplifies DataSet class a bit, adds type hints.

Some code needed to be commented out, but was not used atm

### DIFF
--- a/alphastats/DataSet.py
+++ b/alphastats/DataSet.py
@@ -8,7 +8,7 @@ from alphastats import BaseLoader
 
 
 from alphastats.DataSet_Plot import Plot
-from alphastats.DataSet_Preprocess import Preprocess
+from alphastats.DataSet_Preprocess import Preprocess, PreprocessInterface
 from alphastats.DataSet_Pathway import Enrichment
 from alphastats.DataSet_Statistics import Statistics
 from alphastats.utils import LoaderError
@@ -34,7 +34,7 @@ plotly.io.templates["alphastats_colors"] = plotly.graph_objects.layout.Template(
 plotly.io.templates.default = "simple_white+alphastats_colors"
 
 
-class DataSet(Statistics, Plot, Enrichment):
+class DataSet(PreprocessInterface, Statistics, Plot, Enrichment):
     """Analysis Object"""
 
     def __init__(self, loader, metadata_path=None, sample_column=None):
@@ -83,7 +83,18 @@ class DataSet(Statistics, Plot, Enrichment):
         print("DataSet has been created.")
         self.overview()
 
-    def preprocess(self, **kwargs):
+    def preprocess(
+        self,
+        log2_transform: bool = True,
+        remove_contaminations: bool = False,
+        subset: bool = False,
+        data_completeness: float = 0,
+        normalization: str = None,
+        imputation: str = None,
+        remove_samples: list = None,
+        **kwargs,
+    ):
+        """See documentation in the class implementing PreprocessInterface."""
         pp = Preprocess(
             self.filter_columns,
             self.rawinput,
@@ -94,7 +105,16 @@ class DataSet(Statistics, Plot, Enrichment):
             self.mat,
         )
 
-        self.mat, self.metadata, self.preprocessing_info = pp.preprocess(**kwargs)
+        self.mat, self.metadata, self.preprocessing_info = pp.preprocess(
+            log2_transform,
+            remove_contaminations,
+            subset,
+            data_completeness,
+            normalization,
+            imputation,
+            remove_samples,
+            **kwargs,
+        )
         self.preprocessed = True
 
     def reset_preprocessing(self):
--- a/alphastats/DataSet.py
+++ b/alphastats/DataSet.py
@@ -49,20 +49,21 @@ class DataSet(Statistics, Plot, Enrichment):
 
         """
         self._check_loader(loader=loader)
-        #  load data from loader object
-        self.loader = loader
-        self.rawinput = loader.rawinput
-        self.software = loader.software
-        self.index_column = loader.index_column
-        self.intensity_column = loader.intensity_column
-        self.filter_columns = loader.filter_columns
-        self.evidence_df = loader.evidence_df
-        self.gene_names = loader.gene_names
+
+        self.rawinput: pd.DataFrame = loader.rawinput
+        self.software: str = loader.software
+        self.index_column: str = loader.index_column
+        self.intensity_column: Union[str, list] = loader.intensity_column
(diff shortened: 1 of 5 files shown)


alphastats/DataSet_Preprocess.py
```diff
@@ -9,16 +10,55 @@
 import streamlit as st
 
 from sklearn.experimental import enable_iterative_imputer  # noqa
-from tqdm import tqdm
 
 from alphastats.utils import ignore_warning
 
 
-class Preprocess:
+class PreprocessInterface(ABC):
```
JuliaS92, line 17: Does this have any added benefit besides error catching by the IDE and linters?

To me it looks like it mostly increases complexity and duplicates the definition of the function arguments.

If we use this, I think it should also define, which shared attributes should be present in the child classes, so we have all the documentation of the interface in one place.

alphastats/DataSet_Preprocess.py
```diff
@@ -9,16 +10,55 @@
...
+        self,
+        filter_columns,
+        rawinput,
+        index_column,
+        sample,
+        metadata,
+        preprocessing_info,
+        mat,
+    ):
+        self.filter_columns = filter_columns
```
JuliaS92, line 51: Are these required for preprocessing? Otherwise I would not include them here.

alphastats/DataSet_Preprocess.py
```diff
@@ -215,64 +251,68 @@ def _normalization(self, method: str):
...
 
         data = self.mat.transpose()
         series_of_batches = self.metadata.set_index(self.sample).reindex(
             data.columns.to_list()
         )[batch]
-        self.mat = pycombat(data=data, batch=series_of_batches).transpose()
+
+        batch_corrected_data = pycombat(data=data, batch=series_of_batches).transpose()
+
+        return batch_corrected_data, self.metadata
```
JuliaS92, line 315: self. metadata does not actually change, so it doesn't need to be returned.

alphastats/DataSet_Preprocess.py
```diff
@@ -31,10 +71,6 @@ def _subset(self):
         )
         return self.mat[self.mat.index.isin(self.metadata[self.sample].tolist())]
 
-    def preprocess_print_info(self):
```
JuliaS92, line 34: This is used in the MQ tutorial, so we should replace it there, if we remove it.

alphastats/DataSet.py
```diff
@@ -127,6 +178,15 @@ def _remove_misc_samples_in_metadata(self):
                 "are removed from the metadata."
             )
 
+    # TODO this is implemented in both preprocessing and here
```
JuliaS92, line 181: To fix the todo: This is only needed in the DimensionalityReduction class and only if the step was not run during preprocessing. I would suggest replacing the step in DimensionalityReduction with something like:
`mat = self.data.mat.loc[sample_names,:]` after creating sample_names.

JuliaS92 (commented):
This looks like we're going in the right direciton.
I am not sure if the Interface is maybe overengineering it at the moment. If not I would like to see the shared attributes documented here as well.
We should also try and reduce the number of attributes needed to be passed to Preprocess.
Also I think the metadata actually does not change anytime during batch correction.

alphastats/DataSet_Preprocess.py
mschwoer: they are, in the `_filter` method

alphastats/DataSet_Preprocess.py
mschwoer: it does in `_remove_samples()` (this is the bug that I think this fixes: on `reset_processing`, metadata is not reset)

alphastats/DataSet_Preprocess.py
mschwoer: is it? I couldn't find it anywhere?

alphastats/DataSet_Preprocess.py
mschwoer: I would leave it like this for now, "error catching by the IDE and linters" is alreday quite some value ..also it makes the connection between the implementation and the interface transparent.. 
The shared attributes are already defined: there are none ;-)

mschwoer: update: I realized these are two difference methods .. remove the interface for now

alphastats/DataSet.py
mschwoer: added this as a comment

alphastats/DataSet_Preprocess.py
mschwoer: update: you're right, in `batch_processing`, this is not touched .. thanks for spotting!

alphastats/DataSet_Preprocess.py
JuliaS92: It's an html file, so maybe only in the docs?

JuliaS92 (approved):

alphastats/DataSet_Preprocess.py
mschwoer: removed it also from here https://github.com/MannLabs/alphapeptstats/pull/324/files#diff-46145cbe5211ac920a0dc16d2c095e92dd4730da4dbe6e58776f1b3d407cde09

alphastats/DataSet.py
```diff
@@ -47,38 +54,38 @@ def __init__(self, loader, metadata_path=None, sample_column=None):
...
         if metadata_path is not None:
             self.sample = sample_column
-            self.load_metadata(file_path=metadata_path)
+            self.metadata = self.load_metadata(file_path=metadata_path)
             self._remove_misc_samples_in_metadata()
-
         else:
-            self._create_metadata()
+            self.sample = "sample"
+            self.metadata = pd.DataFrame({"sample": list(self.mat.index)})
```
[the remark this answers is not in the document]
mschwoer: I removed it to have the same level of abstraction in both the `if` and `else` block ..
once there's a reason to re-introduce it, just do it :-)

alphastats/loader/BaseLoader.py
```diff
@@ -98,24 +98,3 @@ def _add_contamination_column(self):
             + "The contaminant library was created by Frankenfield et al."
             + ":https://www.biorxiv.org/content/10.1101/2022.04.27.489766v2.full"
         )
-
-    def read_uploaded_file_into_df(self, file):
```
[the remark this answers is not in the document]
mschwoer: unused code -> remove, unless there's some wisdom in it .. 
you might check `GenericLoader.load_file` which does something similar

PR #324, opened 2024-09-17
Introduce string constants

mschwoer:
introduce string constants for the Preprocessing Keys


alphastats/DataSet_Preprocess.py
```diff
@@ -13,6 +13,28 @@
 from alphastats.utils import ignore_warning
 
 
+class PreprocessingStateKeys:
```
JuliaS92, line 16: I wonder if we can find a shorter name for this. Maybe just PreprocessingStates?
Also this could already be merged with the constants from the preprocessing helper into a separate constants file.
The keys for states that represent whether a step was done could be unified with the key in PreprocessingSteps.

JuliaS92 (approved):
LGTM

alphastats/DataSet_Preprocess.py
mschwoer: I have not merged these with the others yet as I don't know if we can couple those at this stage .. will move and rename in a subsequent PR

PR #325, opened 2024-09-17
move initialization of preprocessing to Preprocess, change initializa…

mschwoer:
move initialization of preprocessing to Preprocess, change initialization of the preprocessing_info

### DIFF
--- a/alphastats/DataSet.py
+++ b/alphastats/DataSet.py
@@ -83,8 +83,15 @@ class DataSet(Statistics, Plot, Enrichment):
             )
             self.intensity_column = intensity_column
 
-        # save preprocessing settings
-        self.preprocessing_info: Dict = self._save_dataset_info()
+        # init preprocessing settings
+        self.preprocessing_info: Dict = Preprocess.init_preprocessing_info(
+            num_samples=self.mat.shape[0],
+            num_protein_groups=self.mat.shape[1],
+            intensity_column=self.intensity_column,
+            filter_columns=self.filter_columns,
+        )
+
+        self.preprocessed = False
         self.preprocessed: bool = False
 
         print("DataSet has been created.")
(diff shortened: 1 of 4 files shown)


alphastats/DataSet.py
```diff
@@ -249,44 +266,15 @@ def load_metadata(self, file_path: Union[pd.DataFrame, str]) -> pd.DataFrame:
...
-            PreprocessingStateKeys.CONTAMINATIONS_REMOVED: False,
-            PreprocessingStateKeys.CONTAMINATION_COLUMNS: self.filter_columns,
-            PreprocessingStateKeys.NUM_REMOVED_PG_DUE_TO_CONTAMINATION: 0,
-            PreprocessingStateKeys.DATA_COMPLETENESS_CUTOFF: 0,
-            PreprocessingStateKeys.NUM_PG_REMOVED_DUE_TO_DATA_COMPLETENESS_CUTOFF: 0,
-            PreprocessingStateKeys.MISSING_VALUES_REMOVED: False,
-        }
-        return preprocessing_dict
-
-    def overview(self):
```
JuliaS92, line 284: I would not delete this. However, I would rename it to __repr__ to follow conventions and add a todo to make this more useful.

alphastats/DataSet.py
```diff
@@ -126,8 +137,16 @@ def preprocess(
...
+        self._create_matrix()
+        self.preprocessing_info = Preprocess.init_preprocessing_info(
+            num_samples=self.mat.shape[0],
+            num_protein_groups=self.mat.shape[1],
+            intensity_column=self.intensity_column,
+            filter_columns=self.filter_columns,
+        )
+
+        self.preprocessed = False
+        # TODO fix bug: metadata is not reset/reloaded here
```
JuliaS92, line 149: I think this would fit well in this branch. If you already addressed this ona nother one, just ignore the comment.

JuliaS92 (approved):
LGTM

TODO: Add a separate attribute in which the settings as passed to preprocess can be stored.

alphastats/DataSet.py
mschwoer: this is not straightforward, so I'd leave it to a later PR

alphastats/DataSet.py
mschwoer: hm, any ideas how to make this more useful?

PR #328, opened 2024-09-17
Decouple dataset from enrichment

mschwoer:
This was easy: Enrichment is not used at all..

### DIFF
--- a/alphastats/DataSet_Pathway.py
+++ b/alphastats/DataSet_Pathway.py
@@ -1,3 +1,4 @@
+# TODO this whole module is unused
 import plotly.express as px
 import requests
 import pandas as pd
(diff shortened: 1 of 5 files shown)


alphastats/DataSet_Pathway.py
```diff
@@ -63,35 +66,21 @@ def plot_bar(self):
...
-                    else:
-                        protein_id = fasta_header_split
-                else:
-                    protein_id = fasta_header_split[1]
-                protein_id_list.append(protein_id)
-            protein_id_concentate = ";".join(protein_id_list)
-            # ADD REV to the protein ID, else there will be duplicates in the ProteinGroup column
-            if "REV_" in entry:
-                protein_id_concentate = "REV_" + protein_id_concentate
+            protein_id_concatenate = AlphaPeptLoader.standardize_protein_group_column(
```
JuliaS92, line 76: This should be part of the data unification executed within the AlphaPeptLoader

JuliaS92 (approved):
Mikhail implemented enrichment code accessible through the interpreter. Maybe some deduplication and merging can be done here in the future.

PR #329, opened 2024-09-17
Decouple dataset from statistics

mschwoer:
This is a huge one, but could not be separated in a sensible way.. 
check the individual commits for the line of reasoning (the bug in fc calculation in [de72c6e](https://github.com/MannLabs/alphapeptstats/pull/329/commits/de72c6e19ce4a459c2581a2ecd57a912e5b3854d) is removed again later

I think now I see where this could be going: the `DataSet` class will contain the data (`mat`, `metadata`) and beside that be just a convenient interface for the frontend. It will offer all operations that can be done on the data (statistics, plots, preprocessing), but delegate the logic to dedicated classes. Two things are missing still:
- decoupling from `Plots`
- moving the logic in `init` (i.e. the actual data loading) to another class (similar to `Statistics`, ..)


alphastats/DataSet_Statistics.py
```diff
@@ -194,36 +130,36 @@ def ancova(
...
-            covariates=covariates,
-            n_permutations=n_permutations,
-            fdr=fdr,
-            s0=s0,
-            subset=subset,
-            plot=True,
-        ).calculate()
-        return res, plot_list
+    # @ignore_warning(RuntimeWarning)
+    # def multicova_analysis(  # TODO never used outside of tests .. how does this relate to multicova.py?
```
JuliaS92, line 134: In MultiCovaAnalysis multicova.py is imported from the calculate function.
```
    def calculate(self):
        from alphastats.multicova import multicova
```

This is defenitely a feature we want in the tool, so it should remain active but we need a ticket for putting it in the interface.

alphastats/gui/utils/analysis_helper.py
```diff
@@ -386,7 +386,3 @@ def load_options():
     st.session_state[StateKeys.STATISTIC_OPTIONS] = statistic_options(st.session_state)
     # TODO: Check if this should be reintroduced or removed
     # st.session_state["interpretation_options"] = interpretation_options
-
-
-def gui_multicova_analysis():
```
JuliaS92, line 391: Add big todo here instead of removing it.

alphastats/plots/VolcanoPlot.py
```diff
@@ -152,7 +175,7 @@ def _sam_calculate_fdr_line(self):
         )
 
     @lru_cache(maxsize=20)
-    def _sam(self):
+    def _sam(self):  # TODO duplicated?
```
JuliaS92, line 178: Only the last two lines differ from the function in DifferentialExpressionAnalysis

JuliaS92 (commented):
LGTM

boopthesnoot (commented):
LGTM

PR #330, opened 2024-09-18
streamline the LLM workflow a bit

mschwoer:
Visually separate LLM config & Analysis, simplify API key setting.

Unfortunately, the red box always appears when trying to access non-existing secrets..

image


JuliaS92 (approved):
LGTM

alphastats/gui/pages/05_LLM.py
```diff
@@ -90,23 +91,27 @@ def select_analysis():
     st.session_state["gpt_submitted_clicked"] = 0
     st.session_state["gpt_submitted_counter"] = 0
 
-c1, c2 = st.columns((1, 2))
 
+st.markdown("#### Configure LLM")
```
JuliaS92, line 95: Maybe make this a card, so it can be hidden away.

mschwoer: then people will forget to do it .. (workflow will be changed anyway soon)

PR #332, opened 2024-09-18
Decouple dataset from plots

mschwoer:
Again, a lot of changes at once. But mostly moved code, and changes to `__init__` methods .. see individual commit, I hope they make sense


alphastats/DataSet_Plot.py
```diff
@@ -364,48 +161,51 @@ def plot_dendrogram(
...
-            plot_df,
-            x="values",
-            color="Imputation",
-            opacity=0.8,
-            hover_data=plot_df.columns,
-            template="simple_white+alphastats_colors",
-        )
-
-        pass
+    # def plot_imputed_values(self):  # TODO not used
```
JuliaS92, line 172: This is a requested feature and should be an issue to add the interface.

alphastats/plots/VolcanoPlot.py
```diff
@@ -33,11 +36,18 @@
 class VolcanoPlot(PlotUtils):
     def __init__(
         self,
-        dataset,
-        group1,
-        group2,
-        column=None,
-        method=None,
+        *,
```
JuliaS92, line 39: Why do we need this?

alphastats/plots/VolcanoPlot.py
```diff
@@ -62,20 +76,39 @@ def __init__(
...
+
         if plot:
             self._perform_differential_expression_analysis()
             self._annotate_result_df()
             self._add_hover_data_columns()
             self._plot()
 
-    # TODO this changes the actual metadata .. is this intended?
-    def _add_metadata_column(self, group1_list: list, group2_list: list):
+    # TODO this used to change the actual metadata .. was this intended?
```
JuliaS92, line 105: Do I get this right:
It used to add the _comparison_column to the DataSet metadata.
Now it returns the altered metadata to VolcanoPlot, which serves the purporse during the Volcano plot generation, but it is never back propagated to the metadata attribute of DataSet?
If that is so, then this is now fine.

JuliaS92 (commented):
I am not sure I like the clutter of DataSet with all the plotting wrappers, especially considering all the feature requests that we have. Could we maybe do the same thing as with the statistics and preprocessing and initialize a plot class once with the data, so we can then do DataSet.plot.Volcano?

alphastats/plots/VolcanoPlot.py
boopthesnoot: it makes people using this class downstream always use keyword arguments when creating an instance in this case

so before it was possible to have a 
```python
VolcanoPlot(mat, rawinput, ...)
```

now it makes us explicitly do 
```python
VolcanoPlot(mat=mat, rawinput=rawinput, ...)
```

boopthesnoot: in terms of __why__ do we need this — probably to prevent errors with the order of positional arguments

alphastats/statistics/DifferentialExpressionAnalysis.py
```diff
@@ -109,10 +109,7 @@ def _sam(self) -> pd.DataFrame:  # TODO duplicated?
 
         transposed = self.mat.transpose()
 
-        if (
-            self.dataset.preprocessing_info[PreprocessingStateKeys.LOG2_TRANSFORMED]
-            is None
-        ):
+        if self.preprocessing_info[PreprocessingStateKeys.LOG2_TRANSFORMED] is None:
```
boopthesnoot, line 112:
```python
if not self.preprocessing_info[PreprocessingStateKeys.LOG2_TRANSFORMED]:
```
do we want to catch only `None`s or do we want to catch all `False`-y values?

boopthesnoot (commented):
👍

alphastats/statistics/DifferentialExpressionAnalysis.py
boopthesnoot: upd: I see that my comment is now outdated

alphastats/DataSet_Plot.py
mschwoer: https://github.com/orgs/MannLabs/projects/12/views/1?filterQuery=multi&pane=issue&itemId=82600738

alphastats/plots/VolcanoPlot.py
mschwoer: exactly. It overwrites the instance variable `self.metadata` with the updated metadata, but does not change the original value.

Note that these instance variables (`mat`, `metadata`, ...) in principle still can be changed in case they are mutable  `pd.DataFrame`, `dict`, ..:
if in `Statistics` something like `self.mat["new_column"] = 123" is done, then it will change the actual `mat` object of the `DataSet` (this fact is used to update `preprocessing_info`)

alphastats/DataSet_Statistics.py
```diff
@@ -17,6 +17,7 @@
 class Statistics:
     def __init__(
         self,
+        *,
```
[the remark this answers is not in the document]
mschwoer: not really, it's just for these very long argument lists .. ideally, only the data that is required for the operations is passed via constructor, and parameters (e.g. for plotting) are passed in the actual `plot` functions..

PR #333, opened 2024-09-18
simplify-dataset-init

mschwoer:
Move some of the init logic to a new class.

Now the DataSet class should contain next to no logic anymore.


alphastats/DataSet.py
```diff
@@ -66,47 +67,81 @@ def __init__(
...
-            self.sample = "sample"
-            self.metadata = pd.DataFrame({"sample": list(self.mat.index)})
-
-        if loader == "Generic":
-            intensity_column = loader._extract_sample_names(
+        self.index_column: str = loader.index_column
+        self.software: str = loader.software
+        self._gene_names: str = loader.gene_names
+
+        self._intensity_column: Union[str, list] = (
```
mschwoer, line 77: note that this was different before:
`loader.intensity_column` was used to create the `mat` object, and only afterwards overwritten in the case of a generic loader ..

alphastats/DataSet.py
```diff
@@ -48,14 +49,14 @@ class DataSet:
...
-        metadata_path: Optional[str] = None,
+        metadata_path: Optional[Union[str, pd.DataFrame]] = None,
         sample_column: Optional[str] = None,
     ):
         """Create DataSet
 
         Args:
             loader (_type_): loader of class AlphaPeptLoader, MaxQuantLoader, DIANNLoader, FragPipeLoader, SpectronautLoader
-            metadata_path (str, optional): path to metadata file. Defaults to None.
+            metadata_path (str or pd.DataFrame, optional): path to metadata file or an actual df. Defaults to None.
```
boopthesnoot, line 59:
```suggestion
            metadata_path_or_df (str or pd.DataFrame, optional): path to metadata file or an actual df. Defaults to None.
```

sorry to be this person hahaha

boopthesnoot (approved):
Thank you! Looks good

alphastats/DataSet.py
mschwoer: this is indeed a good suggestion to make this api better .. in the long run, I'd implement two distinct parameters (`path` and `df`) .. did the whole change here: https://github.com/MannLabs/alphapeptstats/pull/336

JuliaS92:
This looks good to me, wlthough I am not sure the functionalitz provided by the dataset factory really need to/should be separated, as this defines the core of the DataSet.

PR #343, opened 2024-09-20
Fix if condition dev

mschwoer:
same as https://github.com/MannLabs/alphapeptstats/pull/341


boopthesnoot (commented):
Good catch,  thank you!

PR #345, opened 2024-09-27
0 value handling

JuliaS92:
Major changes:
- added raplacement of ==0 values with nan as optional processing step (replacing step in data completeness filter)
- added dropping of none finite features as optional processing step (replacing dropping of all 0 step)
- rewrote the ttest to always operate on log transformed data for fold change and p-value calculation (p-values were calculated in untransformed space if not done durign preprocessing)

Tests were updated accordingly, but subset output changed due to the dropping of values. (Before, rows of nans and 0s were retained). Since this step is also used in batch correction the result changed.

TODO: Figure out which undisplayable p-values make it to the volcano plot function, as the axis of the fold change is wider than it needs to be.


alphastats/DataSet_Preprocess.py
```diff
@@ -334,6 +335,7 @@ def _normalization(self, method: str) -> None:
 
     def _log2_transform(self):
         self.mat = np.log2(self.mat)
+        self.mat = self.mat.replace([np.inf, -np.inf], np.nan)
```
JuliaS92, line 338: TODO: This step should not be necessary if all downstream functions can handle infs.

alphastats/DataSet_Preprocess.py
```diff
@@ -430,6 +434,9 @@ def preprocess(
                 self.mat, self.metadata, self.sample, self.preprocessing_info
             )
 
+        if replace_zero:
```
JuliaS92, line 437: TODO: Write test for this scenario

alphastats/DataSet_Preprocess.py
```diff
@@ -447,9 +454,13 @@ def preprocess(
...
 
-        # TODO should this step be optional, too? is also done in create_matrix
-        # for now, add it to `preprocessing_info`
-        self.mat = self.mat.loc[:, (self.mat != 0).any(axis=0)]
+        self.preprocessing_info.update(
+            {
+                PreprocessingStateKeys.DROP_UNMEASURED_FEATURES: drop_unmeasured_features,
+            }
+        )
+        if drop_unmeasured_features:
```
JuliaS92, line 462: TODO: Write test for this scenario

alphastats/DataSet_Preprocess.py
```diff
@@ -84,6 +85,7 @@ def init_preprocessing_info(
             PreprocessingStateKeys.DATA_COMPLETENESS_CUTOFF: 0,
             PreprocessingStateKeys.NUM_PG_REMOVED_DUE_TO_DATA_COMPLETENESS_CUTOFF: 0,
             PreprocessingStateKeys.MISSING_VALUES_REMOVED: False,
+            PreprocessingStateKeys.DROP_UNMEASURED_FEATURES: False,
```
JuliaS92, line 88: TODO: Add Key for 0 replacement
TODO: Make sure the info is stored in a sensible order for display in the sidebar

alphastats/DataSet_Preprocess.py
boopthesnoot, line 437: To be honest I still don't quite get why we don't unify zeros with nans

alphastats/gui/utils/preprocessing_helper.py
```diff
@@ -177,14 +189,14 @@ def configure_preprocessing(dataset):
         + "[documentation](https://alphapeptstats.readthedocs.io/en/main/data_preprocessing.html)."
     )
 
-    remove_contaminations = st.selectbox(
+    remove_contaminations = st.checkbox(
```
boopthesnoot, line 192: hahaha such a subtle change but it makes much more sense

alphastats/plots/VolcanoPlot.py
```diff
@@ -357,7 +357,9 @@ def _annotate_result_df(self):
         add color labels for up and down regulates
         """
         self.res = self.res[(self.res["log2fc"] < 20) & (self.res["log2fc"] > -20)]
-        self.res["-log10(p-value)"] = -np.log10(self.res[self.pvalue_column])
+        self.res["-log10(p-value)"] = [
```
boopthesnoot, line 360: I don't quite get the change. Isn't it the same, but slower?

alphastats/statistics/DifferentialExpressionAnalysis.py
```diff
@@ -179,15 +187,20 @@ def _ttest(self) -> pd.DataFrame:
         # calculate fold change (if its is not logarithimic normalized)
         mat_transpose = self.mat.transpose()
 
+        if not self.preprocessing_info[PreprocessingStateKeys.LOG2_TRANSFORMED]:
+            mat_transpose = mat_transpose.transform(lambda x: np.log2(x))
```
boopthesnoot, line 191: Should we put up a streamlit message that we log-transformed it under the hood? Otherwise, it isn't transparent and won't be reproducible in a standalone analysis without looking at the source code

alphastats/DataSet_Preprocess.py
JuliaS92: There are instances, where the uploaded data might include real 0 values. For example in one of my own usecases where I would express profiles as percentages and then impute with 0. This would happen outside upfront and be a custom upload with real 0s.

alphastats/plots/VolcanoPlot.py
JuliaS92: With the changes to the upstream functions, masked p-values emerge (automatic completeness filtering). Since self.res["-log10(p-value)"] is a pandas Series, it tries to use the pandas log10 function even though we write np.log10. The pandas function can't handle the masked values though (log10 0 error) and the only way I found to handle them was to iterate manually.

alphastats/statistics/DifferentialExpressionAnalysis.py
JuliaS92: I would make it transparent by the axis labelling and as static text on the interface (calling back to streamlit from here would be a hassle). In the long run we can make it optional, in case someone wants to run t-tests on already normally distribtued data.

alphastats/statistics/DifferentialExpressionAnalysis.py
```diff
@@ -231,17 +244,13 @@ def _pairedttest(self) -> pd.DataFrame:
         df["log2fc"] = fc
         return df
 
-    def _calculate_foldchange(  # TODO duplicated
+    def _calculate_logfoldchange(  # TODO duplicated
         self, mat_transpose: pd.DataFrame, group1_samples: list, group2_samples: list
     ):
         group1_values = mat_transpose[group1_samples].T.mean().values
```
mschwoer, line 250: please add a docstring making clear that this method assumes log-transformed data

tests/gui/test_04_preprocessing.py
```diff
@@ -28,7 +28,7 @@ def test_page_04_loads_with_input():
     at.run()
 
     assert not at.exception
-    assert at.columns[3].selectbox.len == 6
+    assert at.columns[3].selectbox.len == 3
     assert at.button.len == 2
```
mschwoer, line 32: add a check for the newly added check boxes?

alphastats/DataSet_Preprocess.py
```diff
@@ -447,9 +454,13 @@ def preprocess(
...
             self._imputation(method=imputation)
 
-        # TODO should this step be optional, too? is also done in create_matrix
-        # for now, add it to `preprocessing_info`
-        self.mat = self.mat.loc[:, (self.mat != 0).any(axis=0)]
+        self.preprocessing_info.update(
+            {
+                PreprocessingStateKeys.DROP_UNMEASURED_FEATURES: drop_unmeasured_features,
+            }
+        )
```
mschwoer, line 461: please move this into the `if drop_unmeasured_features:` condition

alphastats/plots/VolcanoPlot.py
JuliaS92: I added a todo comment.

alphastats/statistics/DifferentialExpressionAnalysis.py
JuliaS92: Will do in a future pr going over the whole range of docstrings

tests/gui/test_04_preprocessing.py
JuliaS92: This was just to check any interface is loaded. I don't want to tie down the exact widgets too strictly.

mschwoer (approved):

PR #349, opened 2024-10-18
liniting_rule_F

mschwoer:
fixes for linting rule F


JuliaS92 (approved):
Looks good.
I think we will need to put exact values on all the volcano tests, once we verified the functions.

PR #350, opened 2024-10-18
liniting_rule_E

mschwoer:
fixes for linting rules E
get rid of the non-breaking spaces


alphastats/gui/utils/analysis_helper.py
```diff
@@ -37,7 +37,7 @@ def save_plot_to_session_state(plot, method):
 
 
 def display_df(df):
```
JuliaS92, line 39: This should use pandas stylers, rather than changing the data, as this can become quite a lot to handle. Also not sure, why we need this.

alphastats/utils.py
```diff
@@ -73,34 +73,35 @@ def inner(*args, **kwargs):
...
-        result = tuple(result) if type(result) == list else result
+        result = tuple(result) if isinstance(result, list) else result
         return result
 
     return wrapper
 
 
-def find_duplicates_in_list(l: list) -> list:
-    """Find duplicates in a list
+def find_duplicates_in_list(input_list: list) -> list:
```
JuliaS92, line 92: This function is only used once to assert index uniqueness, and should be replaced with https://pandas.pydata.org/docs/reference/api/pandas.Index.has_duplicates.html#pandas.Index.has_duplicates

JuliaS92 (approved):

PR #351, opened 2024-10-18
simplify_analysis-II

mschwoer:
simplify option loading for analysis page

### DIFF
--- a/alphastats/gui/pages/04_Analysis.py
+++ b/alphastats/gui/pages/04_Analysis.py
@@ -1,4 +1,5 @@
 import streamlit as st
+from gui.utils.options import get_plotting_options, get_statistic_options
 
 from alphastats.gui.utils.analysis_helper import (
     display_df,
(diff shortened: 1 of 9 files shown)


boopthesnoot (approved):
good decluttering effort!

alphastats/gui/pages/04_Analysis.py
```diff
@@ -1,12 +1,12 @@
 import streamlit as st
+from gui.utils.options import get_plotting_options, get_statistic_options
```
JuliaS92, line 2: I get that we wanted to declutter the session state. How expensive is it to run these functions? Maybe we can at least restrict that to once per page reload, or cache the result.

JuliaS92 (approved):
Since we wanted to prioritize function over performance now, my one comment is mostly for future consideration.

alphastats/gui/pages/04_Analysis.py
mschwoer: they will be removed in a later PR

PR #353, opened 2024-10-21
Refactor llm i

mschwoer:
Note that this is not the final state in terms of UI/UX, but just a step towards fully integrating LLM.

- reduce code duplication for LLM part by reusing the information generated in Analysis.
- simplify logic on LLM page
- remove some keys from session state

### DIFF
--- a/alphastats/gui/pages/05_LLM.py
+++ b/alphastats/gui/pages/05_LLM.py
@@ -42,6 +42,8 @@ def llm_config():
         if st.session_state[StateKeys.API_TYPE] == "gpt4o":
             api_key = st.text_input("Enter OpenAI API Key", type="password")
             set_api_key(api_key)
+        else:
+            st.info("Expecting Ollama API at http://localhost:11434.")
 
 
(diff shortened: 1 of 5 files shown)


alphastats/gui/pages/05_LLM.py
```diff
@@ -17,171 +14,64 @@
...
-        StateKeys.PLOT_SUBMITTED_CLICKED
-    ]
-    volcano_plot = gui_volcano_plot_differential_expression_analysis(
-        chosen_parameter_dict
-    )
-    volcano_plot._update(plotting_parameter_dict)
-    volcano_plot._annotate_result_df()
-    volcano_plot._plot()
-    genes_of_interest_colored = volcano_plot.get_colored_labels()
+    # TODO move this to volcano anyway ?
```
JuliaS92, line 63: These should ideally be functions of all analyses. To get a list of all significant proteins and be able to look them up on uniprot.

JuliaS92 (approved):

PR #355, opened 2024-10-22
Refactor llm iv

mschwoer:
Mostly moving code around.


alphastats/llm/llm_integration.py
```diff
@@ -186,25 +192,29 @@ def execute_function(
...
                 return function(**function_args)
 
-            # special treatment for this one
+            # special treatment for these functions
+            elif function_name == "perform_dimensionality_reduction":
+                # TODO add API in dataset
+                perform_dimensionality_reduction(self._dataset, **function_args)
+
             elif function_name == "plot_intensity":
+                # TODO move this logic to dataset
```
JuliaS92, line 205: This relates to my comment on the previous pull request. This logic is not needed, if protein ids are passed, as it was (by name) the case before refactor llm iii.

JuliaS92 (approved):

alphastats/llm/llm_integration.py
mschwoer: is moved to dataset later

PR #357, opened 2024-10-23
Refactor llm vi

mschwoer:
- add "test api" functionality
- allow multiple models in parallel
- add support for a smaller ollama for local development
- add some instructions to readme

### DIFF
--- a/alphastats/gui/pages/05_LLM.py
+++ b/alphastats/gui/pages/05_LLM.py
@@ -10,10 +10,14 @@ from alphastats.gui.utils.analysis_helper import (
 from alphastats.gui.utils.llm_helper import (
     display_proteins,
     set_api_key,
+    test_llm_connection,
 )
 from alphastats.gui.utils.ui_helper import StateKeys, init_session_state, sidebar_info
 from alphastats.llm.llm_integration import LLMIntegration, Models
 from alphastats.llm.prompts import get_initial_prompt, get_system_message
+
+OLLAMA_BASE_URL = os.getenv("OLLAMA_BASE_URL", "http://localhost:11434")
+
 
 init_session_state()
 sidebar_info()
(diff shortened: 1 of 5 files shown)


alphastats/gui/pages/05_LLM.py
```diff
@@ -162,6 +192,11 @@ def llm_config():
 def llm_chat(llm_integration: LLMIntegration, show_all: bool = False):
     """The chat interface for the LLM analysis."""
 
+    # TODO dump to file -> static file name, plus button to do so
```
JuliaS92, line 195: I say save chat as txt, without encoding objects, just put a replacement string. Offer bulk download of zip with all figures (via plotly download as svg.). Alternatively write it all in one pdf report using e.g. pdfrw and reportlab (I have code for that combo).

alphastats/gui/pages/05_LLM.py
```diff
@@ -162,6 +192,11 @@ def llm_config():
...
+    # how to deal with binaries? base64 encode?
+    # "import chat" functionality?
```
JuliaS92, line 197: I think that would be overkill. If we have a txt export, one could always sopy that into a prompt oneself and tell the LLM that this was the previous session.

alphastats/llm/llm_integration.py
```diff
@@ -28,8 +28,9 @@
 
 
 class Models:
-    GPT = "gpt-4o"
-    OLLAMA = "llama3.1:70b"
+    GPT4O = "gpt-4o"
+    OLLAMA_31_8B = "llama3.1:8b"
+    OLLAMA_31_70B = "llama3.1:70b"  # for testing only
```
JuliaS92, line 33: Comment on the wrong line?

JuliaS92 (approved):

alphastats/llm/llm_integration.py
mschwoer: good catch :-)

PR #358, opened 2024-10-23
Refactor llm vii

mschwoer:
- make link between LLM function definitions and actual functions transparent
- add general dimensionality reduction to DataSet
- simplify some helpers
- slightly refactor llm_integration
- add unit tests for whole LLM module

### DIFF
--- a/alphastats/DataSet.py
+++ b/alphastats/DataSet.py
@@ -311,6 +311,27 @@ class DataSet:
         )
         return dimensionality_reduction.plot
 
+    def perform_dimensionality_reduction(
+        self, method: str, group: Optional[str] = None, circle: bool = False
+    ):
+        """Generic wrapper for dimensionality reduction methods to be used by LLM.
+
+        Args:
+            method (str): "pca", "tsne", "umap"
+            group (str, optional): column in metadata that should be used for coloring. Defaults to None.
+            circle (bool, optional): draw circle around each group. Defaults to False.
+        """
+
+        result = {
+            "pca": self.plot_pca,
+            "tsne": self.plot_tsne,
+            "umap": self.plot_umap,
+        }.get(method)
+        if result is None:
+            raise ValueError(f"Invalid method: {method}")
+
+        return result(group=group, circle=circle)
+
     @ignore_warning(RuntimeWarning)
     def plot_volcano(
         self,
--- a/alphastats/gui/pages/05_LLM.py
+++ b/alphastats/gui/pages/05_LLM.py
@@ -64,11 +64,10 @@ def llm_config():
                     api_key=st.session_state[StateKeys.OPENAI_API_KEY],
                     base_url=base_url,
                 )
-            st.success(
-                f"Connection to {api_type} successful!"
-            ) if error is None else st.error(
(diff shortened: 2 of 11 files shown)


alphastats/gui/pages/05_LLM.py
```diff
@@ -36,11 +36,15 @@ def llm_config():
...
+        current_model = st.session_state.get(StateKeys.MODEL_NAME, None)
 
+        models = [Models.GPT4O, Models.OLLAMA_31_70B, Models.OLLAMA_31_8B]
         st.session_state[StateKeys.MODEL_NAME] = st.selectbox(
             "Select LLM",
-            [Models.GPT4O, Models.OLLAMA_31_70B, Models.OLLAMA_31_8B],
+            models,
+            index=models.index(st.session_state.get(StateKeys.MODEL_NAME))
+            if current_model is not None
+            else 0,
```
JuliaS92, line 47: key = StateKeys.MODEL_NAME?

alphastats/gui/pages/05_LLM.py
```diff
@@ -67,9 +71,9 @@ def llm_config():
                 if error is None:
                     st.success(f"Connection to {model_name} successful!")
                 else:
-                    st.error(f"❌ Connection to {model_name} failed: {str(error)}")
+                    st.error(f"Connection to {model_name} failed: {str(error)}")
 
-        if model_before != st.session_state[StateKeys.MODEL_NAME]:
+        if current_model != st.session_state[StateKeys.MODEL_NAME]:
```
JuliaS92, line 76: If this is the intendend behaviour, then why is the config a st.fragment?

alphastats/DataSet.py
```diff
@@ -311,6 +311,27 @@ def plot_umap(self, group: Optional[str] = None, circle: bool = False):
...
+
+        result = {
+            "pca": self.plot_pca,
+            "tsne": self.plot_tsne,
+            "umap": self.plot_umap,
+        }.get(method)
+        if result is None:
+            raise ValueError(f"Invalid method: {method}")
+
+        return result(group=group, circle=circle)
```
JuliaS92, line 333: It would be great to have something similarly simple for the differential analysis longterm.

tests/llm/test_llm_utils.py
```diff
@@ -0,0 +1,100 @@
...
+def test_get_protein_id_empty_map():
+    """Test with empty mapping dictionary."""
+    result = get_protein_id_for_gene_name("VCL", {})
+    assert result == "VCL"
+
+
+def test_get_protein_id_multiple_matches(gene_to_prot_map):
+    """Test with a gene that appears in multiple compound keys."""
+    result = get_protein_id_for_gene_name("MULTI", gene_to_prot_map)
+    assert result == "PROT1;PROT2;PROT3"
```
JuliaS92, line 94: I think this is the same as test_get_protein_id_compound_key. Actually VCL would be the one matching multiple protein ids.

JuliaS92 (approved):
I think we need to have a conversation about genes vs proteins with everyone and decide on one to use (i am in favor of protein ids) in the backend. Display can be different, but should match all cases. Otherwise kudos :)

PR #360, opened 2024-10-24
Refactor volcano ii

mschwoer:
simplify volcano implementation and call from analysis page

### DIFF
--- a/alphastats/gui/pages/05_LLM.py
+++ b/alphastats/gui/pages/05_LLM.py
@@ -93,7 +93,9 @@ st.write(f"Parameters used for analysis: {parameter_dict}")
 c1, c2 = st.columns((1, 2))
 
 with c1:
-    genes_of_interest_df = volcano_plot.get_colored_labels_df()
+    genes_of_interest_df = volcano_plot.res
+    genes_of_interest_df = genes_of_interest_df[genes_of_interest_df["label"] != ""]
+
     gene_names_colname = st.session_state[StateKeys.LOADER].gene_names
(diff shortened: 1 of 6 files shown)


alphastats/gui/pages/05_LLM.py
```diff
@@ -93,7 +93,9 @@ def llm_config():
 c1, c2 = st.columns((1, 2))
 
 with c1:
-    genes_of_interest_df = volcano_plot.get_colored_labels_df()
+    genes_of_interest_df = volcano_plot.res
+    genes_of_interest_df = genes_of_interest_df[genes_of_interest_df["label"] != ""]
```
JuliaS92, line 97: 1 instead of 2 lines?

alphastats/plots/ClusterMap.py
```diff
@@ -62,7 +63,7 @@ def _prepare_df(self):
             significant_proteins = anova_df[anova_df["ANOVA_pvalue"] < 0.05][
                 self.index_column
             ].to_list()
-            df = df[significant_proteins]
+            df = df[significant_proteins]  # TODO bug?
```
JuliaS92, line 66: No, this is intended, however it would be better if the statistical results could be selected here, instead of running anova (intransparently).

JuliaS92 (approved):

alphastats/plots/ClusterMap.py
mschwoer: I was referring to the fact that `df` is not used after this line ..

PR #361, opened 2024-10-24
Refactor volcano iii

mschwoer:
move the gene to protein id mapping to DataSet, thus simplifying LLM part

### DIFF
--- a/alphastats/DataSet.py
+++ b/alphastats/DataSet.py
@@ -441,6 +441,7 @@ class DataSet:
 
     def plot_intensity(
         self,
+        *,
         protein_id: str = None,
         gene_name: str = None,
         group: str = None,
--- a/alphastats/DataSet.py
+++ b/alphastats/DataSet.py
@@ -94,6 +94,17 @@ class DataSet:
         self.sample: str = sample
         self.preprocessing_info: Dict = preprocessing_info
 
+        self.gene_name_to_protein_id_map = (
+            dict(
+                zip(
+                    self.rawinput[self._gene_names].tolist(),
+                    self.rawinput[self.index_column].tolist(),
+                )
+            )
+            if self._gene_names
+            else {}
+        )
+
         print("DataSet has been created.")
 
     def _get_init_dataset(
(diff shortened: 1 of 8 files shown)


alphastats/DataSet.py
```diff
@@ -94,6 +94,21 @@ def __init__(
...
 
+        self._gene_name_to_protein_id_map = (
+            {
+                k: v
+                for k, v in dict(
+                    zip(
+                        self.rawinput[self._gene_names].tolist(),
+                        self.rawinput[self.index_column].tolist(),
+                    )
+                ).items()
```
JuliaS92, line 105: This is not necessarily unique, and should ideally raise an error in some of our test-data sets that contain isoform ids. E.g. TPM1 occurs 5 times in testfiles/maxquant/proteinGroups.txt with different base Protein IDs.

tests/test_DataSet.py
```diff
@@ -776,6 +776,18 @@ def test_multicova_analysis_invalid_covariates(self):
...
+            self.obj._get_protein_id_for_gene_name("MADE_UP_GENE"), "MADE_UP_GENE"
+        )
+        self.assertEqual(
+            self.obj._get_protein_id_for_gene_name("ALDOC"),
+            "P09972;A0A024QZ64;A8MVZ9;B7Z3K9;B7Z1N6;B7Z3K7;J3KSV6;J3QKP5;C9J8F3;B7Z1Z9;J3QKK1;B7Z1H6;K7EKH5;B7Z1L5",
+        )
+        self.assertEqual(
+            self.obj._get_protein_id_for_gene_name("FCGRT"), "P55899;M0R0A9;A0A024QZI2"
+        )
+
```
JuliaS92, line 790: TODO: This needs to cover cases akin to what has been deleted, and needs to be tested on an artificial frame, as different searche engine outputs cover different cases.

JuliaS92 (approved):
LGTM
Anything regarding ids needs to be addressed in a separate PR.

tests/test_DataSet.py
mschwoer: not sure what you mean .. all the branches of the method are tested .. a artifical frame would make the test more easy to read, but that's out current testing strategy ;-)

PR #364, opened 2024-10-28
Remove sample column

mschwoer:
- harmonize the "sample" column
- simplify session state

### DIFF
--- a/alphastats/DataSet.py
+++ b/alphastats/DataSet.py
@@ -64,19 +64,20 @@ class DataSet:
         """
         self._check_loader(loader=loader)
 
+        self._data_harmonizer = DataHarmonizer(loader, sample_column)
+
         # fill data from loader
-        self.rawinput: pd.DataFrame = DataHarmonizer(loader).get_harmonized_rawinput(
+        self.rawinput: pd.DataFrame = self._data_harmonizer.get_harmonized_rawinput(
             loader.rawinput
         )
         self.filter_columns: List[str] = loader.filter_columns
-
         self.software: str = loader.software
-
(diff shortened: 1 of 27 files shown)


alphastats/DataSet.py
```diff
@@ -64,19 +64,20 @@ def __init__(
         """
         self._check_loader(loader=loader)
 
+        self._data_harmonizer = DataHarmonizer(loader, sample_column)
```
JuliaS92, line 67: I would be in favor of moving the data harmonization out of dataset and insert it as last step in the loaders instead. This way we could immediately pass the harmonized data as rawinput and remove the loader from the arguments. The loader could then be a transient object, removing one instance of data duplication.

alphastats/DataSet_Plot.py
```diff
@@ -95,15 +94,15 @@ def plot_sampledistribution(
         # create long df
         matrix = self.mat if not use_raw else self.rawmat
         df = matrix.unstack().reset_index()
-        df.rename(columns={"level_1": self.sample, 0: "Intensity"}, inplace=True)
+        df.rename(columns={"level_1": Cols.SAMPLE, 0: "Intensity"}, inplace=True)
```
JuliaS92, line 97: Seeing this here, but I think we should replace intensity either with the more generic term abundance, or use what was actually the original name. Intensity or LFQ intensity, or even SILAC ratio makes a bit difference.

JuliaS92 (approved):

alphastats/DataSet.py
mschwoer: agree, still I followed the "dont touch the loaders for now" principle :-)

PR #365, opened 2024-10-29
Refactor llm iii

mschwoer:
- decouple get_gene_to_prot_id_mapping from session state
- fix bug when calling DimensionalityReduction
- refactor function calling in LLM

### DIFF
--- a/alphastats/gui/utils/gpt_helper.py
+++ b/alphastats/gui/utils/gpt_helper.py
@@ -305,7 +305,7 @@ def perform_dimensionality_reduction(group, method, circle, **kwargs):
     return dr.plot
 
 
-def get_gene_to_prot_id_mapping(
+def get_protein_id_for_gene_name(
     gene_name: str, gene_to_prot_id_map: Dict[str, str]
 ) -> str:
     """Get protein id from gene id. If gene id is not present, return gene id, as we might already have a gene id.
--- a/alphastats/gui/utils/gpt_helper.py
+++ b/alphastats/gui/utils/gpt_helper.py
@@ -1,4 +1,3 @@
-import copy
 from typing import Dict, List
 
 import pandas as pd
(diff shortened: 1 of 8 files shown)


alphastats/gui/utils/ollama_utils.py
```diff
@@ -179,24 +181,39 @@ def execute_function(
             If the function is not implemented or the dataset is not available
         """
         try:
-            if function_name == "get_gene_function":
-                # TODO log whats going on
```
JuliaS92, line 183: What happened to this TODO?

alphastats/gui/utils/ollama_utils.py
```diff
@@ -179,24 +181,39 @@ def execute_function(
...
+            elif function_name == "plot_intensity":
+                gene_name = function_args.pop("gene_name")
+                protein_id = get_gene_to_prot_id_mapping(
+                    gene_name, self._gene_to_prot_id_map
+                )
+                function_args["protein_id"] = protein_id
+
+                return self.dataset.plot_intensity(**function_args)
+
+            # fallback: try to find the function in the Dataset functions
```
JuliaS92, line 204: This whole logic is does not work as well anymore, now that we refactored most methods to no longer be directly available from dataset.

alphastats/gui/utils/gpt_helper.py
```diff
@@ -153,21 +149,21 @@ def get_assistant_functions(
                 "parameters": {
                     "type": "object",
                     "properties": {
-                        "protein_id": {
+                        "gene_name": {  # this will be mapped to "protein_id" when calling the function
```
JuliaS92, line 152: Have you actually tested this? I'm not sure whether you also changed the corresponding function to use gene names instead of protein ids, or the function was wrong before. It should be clear in both instances if it's actually using protein ids or gene names. Generally we should always use protein ids internally, as 1. gene names can change over time, 2. not all uploaded datasets contain gene names.

JuliaS92 (approved):
LGTM, but please take a look at the comments.

alphastats/gui/utils/gpt_helper.py
JuliaS92: https://github.com/MannLabs/alphapeptstats/pull/355/files#r1834408990
Found this in the next pr

alphastats/gui/utils/ollama_utils.py
mschwoer: there's now the option to show the whole conversation

alphastats/gui/utils/ollama_utils.py
mschwoer: note sure what you mean? all method calls should still work

alphastats/gui/utils/gpt_helper.py
mschwoer: yes, there's some back-and-forth with this mapping ;-)

PR #366, opened 2024-10-30
Refactor analysis i

mschwoer:
introduce an `Analysis` class that handles analysis parameters selection and mapping to the respective methods,
only for Volcano for now.

plus unrelated changes(TM):
- make plots show on the right
- remove biopython dependency and filling of Bio.Entrez

### DIFF
--- a/alphastats/gui/gui.py
+++ b/alphastats/gui/gui.py
@@ -18,6 +18,8 @@ def run():
     os.system(
         f"python -m streamlit run AlphaPeptStats.py --global.developmentMode=false {extra_args}"
     )
+
+    # TODO why are we starting the app a second time here?
     _this_file = os.path.abspath(__file__)
     _this_directory = os.path.dirname(_this_file)
 
--- a/alphastats/gui/utils/analysis.py
+++ b/alphastats/gui/utils/analysis.py
@@ -1,0 +1,148 @@
+"""Module providing frontend widgets for gathering parameters and mapping them to the actual analysis."""
+
+from abc import ABC, abstractmethod
+from collections import defaultdict
+
+import streamlit as st
+
+from alphastats.keys import Cols
+from alphastats.plots.VolcanoPlot import VolcanoPlot
+
+
+class Analysis(ABC):
+    """Abstract class for analysis widgets."""
+
+    def __init__(self, dataset):
+        self._dataset = dataset
+        self._parameters = defaultdict(lambda: None)
+
+    @abstractmethod
+    def show_widget(self):
+        """Show the widget and gather parameters."""
+        pass
+
+    @abstractmethod
+    def do_analysis(self):
+        """Perform the analysis."""
+        pass
+
+
+class GroupCompareAnalysis(Analysis, ABC):
+    """Abstract class for group comparison analysis widgets."""
+
+    def show_widget(self):
+        """Gather paramrters to compare two group."""
+
+        metadata = self._dataset.metadata
+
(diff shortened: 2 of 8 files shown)


alphastats/gui/utils/analysis.py
```diff
@@ -0,0 +1,148 @@
...
+        pass
+
+
+class GroupCompareAnalysis(Analysis, ABC):
+    """Abstract class for group comparison analysis widgets."""
+
+    def show_widget(self):
+        """Gather paramrters to compare two group."""
+
+        metadata = self._dataset.metadata
```
mschwoer, line 36: the logic is copied from `helper_compare_two_groups`

alphastats/gui/utils/analysis.py
```diff
@@ -0,0 +1,155 @@
...
+        grouping_variable = st.selectbox(
+            "Grouping variable",
+            options=[default_option]
+            + metadata.columns.to_list()
+            + [custom_group_option],
+        )
+
+        column = None
+        if grouping_variable == default_option:
+            st.stop()  # TODO: using stop here is not really great
```
boopthesnoot, line 55: Why? Because we might some logic down the line?

alphastats/gui/gui.py
```diff
@@ -18,6 +18,8 @@ def run():
     os.system(
         f"python -m streamlit run AlphaPeptStats.py --global.developmentMode=false {extra_args}"
     )
+
+    # TODO why are we starting the app a second time here?
```
boopthesnoot, line 22: good question! weird thing, I checked and the functionality did not seem to chage

boopthesnoot (approved):

alphastats/gui/utils/analysis.py
JuliaS92: I think we should not stop here. Otherwise if might break other part of the page, that would only be shown afterwards.

alphastats/gui/utils/analysis.py
```diff
@@ -0,0 +1,155 @@
...
+
+        parameters["min_fc"] = st.select_slider(
+            "Foldchange cutoff", range(0, 3), value=1
+        )
+
+        if method == "sam":
+            parameters["perm"] = st.number_input(
+                label="Number of Permutations", min_value=1, max_value=1000, value=10
+            )
+            parameters["fdr"] = st.number_input(
```
JuliaS92, line 122: The sam fdr cutoff should be mutually exclusive with alpha

alphastats/gui/utils/analysis.py
```diff
@@ -0,0 +1,155 @@
...
+                label="FDR cut off", min_value=0.005, max_value=0.1, value=0.050
+            )
+
+        self._parameters.update(parameters)
+
+    def do_analysis(self):
+        """Draw Volcano Plot using the VolcanoPlot class.
+
+        Returns a tuple(figure, analysis_object, parameters) where figure is the plot,
+        analysis_object is the underlying object, parameters is a dictionary of the parameters used.
```
JuliaS92, line 132: Not sure if this will be addessed in the other unreviewed PRs, but I would think a split into calculating analysis and plotting results.

JuliaS92 (approved):

alphastats/gui/utils/analysis.py
mschwoer: this was resolved when adressing your feedback on the webapp (to always show all input fields)

PR #367, opened 2024-10-30
Refactor analysis ii

mschwoer:
refactor dim reduction analysis

### DIFF
--- a/alphastats/gui/pages/04_Analysis.py
+++ b/alphastats/gui/pages/04_Analysis.py
@@ -1,11 +1,15 @@
 import streamlit as st
 
+from alphastats.gui.utils.analysis import PlottingOptions
 from alphastats.gui.utils.analysis_helper import (
     display_df,
     display_plot,
     do_analysis,
 )
-from alphastats.gui.utils.options import get_plotting_options, get_statistic_options
+from alphastats.gui.utils.options import (
+    get_plotting_options,
+    get_statistic_options,
+)
 from alphastats.gui.utils.ui_helper import (
     StateKeys,
     convert_df,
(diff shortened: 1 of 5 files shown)


alphastats/gui/pages/04_Analysis.py
```diff
@@ -44,18 +48,27 @@
...
+    plotting_options = PlottingOptions.get_values()
     method = st.selectbox(
         "Analysis",
         options=["<select>"]
         + ["------- plots -------"]
-        + list(get_plotting_options(st.session_state).keys())
+        + plotting_options
+        + [
+            key
+            for key in list(get_plotting_options(st.session_state).keys())
```
boopthesnoot, line 59: dumb question: why do we have a separate `plotting_options` thing again? why not just have the output of `get_plotting_options`? because we anyways have both here?

alphastats/plots/DimensionalityReduction.py
```diff
@@ -119,7 +120,7 @@ def _pca(self):
         }
 
     def _tsne(self, **kwargs):
-        tsne = sklearn.manifold.TSNE(n_components=2, verbose=1, **kwargs)
```
boopthesnoot, line 122: why? :)

why not go from `sklearn.decomposition.PCA` to `PCA` too in that case? (just a question, I actually like verbose init because it's instantly clear from which package the class comes)

boopthesnoot (approved):
LGTM

alphastats/gui/pages/04_Analysis.py
boopthesnoot: UPD: I see it's gone in the next one

alphastats/plots/DimensionalityReduction.py
mschwoer: agree, will revert to verbose import in this case

JuliaS92 (approved):

PR #368, opened 2024-10-30
Refactor analysis iii

mschwoer:
refactor analysis of clustermap, sampledistribution and intensity plot

### DIFF
--- a/alphastats/gui/pages/04_Analysis.py
+++ b/alphastats/gui/pages/04_Analysis.py
@@ -7,7 +7,6 @@ from alphastats.gui.utils.analysis_helper import (
     do_analysis,
 )
 from alphastats.gui.utils.options import (
-    get_plotting_options,
     get_statistic_options,
 )
 from alphastats.gui.utils.ui_helper import (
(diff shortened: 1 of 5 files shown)


alphastats/gui/pages/04_Analysis.py
```diff
@@ -54,21 +53,13 @@
...
-        ]
         + ["------- statistics -------"]
         + list(get_statistic_options(st.session_state).keys()),
     )
 
-    if method in (
-        list((plot_options := get_plotting_options(st.session_state)).keys())
-        + plotting_options
-    ):
+    if method in (plotting_options):
```
boopthesnoot, line 60:
```suggestion
    if method in plotting_options:
```
is there a hidden reason?

boopthesnoot (approved):
LGTM

alphastats/gui/pages/04_Analysis.py
mschwoer: nope, just slipped through ;-)

PR #370, opened 2024-10-30
Refactor analysis v

mschwoer:
refactor anova & ancova analysis

### DIFF
--- a/alphastats/gui/pages/04_Analysis.py
+++ b/alphastats/gui/pages/04_Analysis.py
@@ -6,9 +6,6 @@ from alphastats.gui.utils.analysis_helper import (
     display_plot,
     do_analysis,
 )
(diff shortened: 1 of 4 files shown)


boopthesnoot (approved):

alphastats/gui/pages/04_Analysis.py
```diff
@@ -35,6 +32,8 @@
     """
 st.markdown(styl, unsafe_allow_html=True)
 
+# TODO put everything in the session state for a given parameter set?
+# or is caching functionality the way to go here?
```
JuliaS92, line 36: Caching is the way to go. But only chache calculations, not plots. The expensive part about plots is the rendering in the browser and we can't cache that away.

JuliaS92 (approved):

PR #371, opened 2024-10-30
Refactor analysis vi

mschwoer:
complete the analysis refactoring by simplifying & (re)moving code

add new feature: download plotting parameters and show them on the "results" page

### DIFF
--- a/alphastats/gui/pages/06_Results.py
+++ b/alphastats/gui/pages/06_Results.py
@@ -21,9 +21,11 @@ for count, saved_item in enumerate(st.session_state[StateKeys.PLOT_LIST]):
 
     method = saved_item[0]
     plot = saved_item[1]
+    parameters = saved_item[2]
 
     st.markdown("\n\n\n")
     st.markdown(f"#### {method}")
+    st.write(f"Parameters used for analysis: {parameters}")
 
     if st.button("x remove analysis", key="remove" + method + str(count)):
         st.session_state[StateKeys.PLOT_LIST].remove(saved_item)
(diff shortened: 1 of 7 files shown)


alphastats/gui/pages/06_Results.py
```diff
@@ -21,9 +21,11 @@
 
     method = saved_item[0]
```
boopthesnoot, line 22:
```suggestion
for count, (method, plot, parameters) in enumerate(st.session_state[StateKeys.PLOT_LIST]):
```
we could change the loop to this, or is it making this less readable?

alphastats/gui/utils/analysis.py
```diff
@@ -58,22 +46,22 @@ def show_widget(self):  # noqa: B027
         pass
 
     def do_analysis(self):
-        """Perform the analysis.
+        """Perform the analysis after an optional check for NaNs.
```
boopthesnoot, line 49: i didn't understand what you meant by saying that this check is optional?

alphastats/gui/utils/analysis.py
```diff
@@ -449,4 +448,20 @@ def _do_analysis(self):
             covar=self._parameters["covar"],
             between=self._parameters["between"],
         )
-        return ancova_analysis, None, self._parameters
+        return ancova_analysis, None
+
+
+ANALYSIS_OPTIONS = {
```
boopthesnoot, line 454: would moving the constants to the beggining of the file make more sense?

boopthesnoot (approved):
LGTM

alphastats/gui/pages/06_Results.py
mschwoer: in principle, yes, but then the `.remove(saved_item)` would become less readable

alphastats/gui/utils/analysis.py
mschwoer: will rephrase it to `Perform the analysis after some upfront method-dependent checks (e.g. or NaNs)`

alphastats/gui/utils/analysis.py
mschwoer: totally, but this needs to be done after the class definitions referenced in the dict

PR #373, opened 2024-10-31
Refactor analysis viii

mschwoer:
All the quickwins that we discussed today, see individual commits for details.


boopthesnoot (approved):
LGTM

alphastats/gui/utils/analysis.py
```diff
@@ -46,19 +50,39 @@ def show_widget(self):  # noqa: B027
...
     @abstractmethod
-    def _do_analysis(self):
+    def _do_analysis(
+        self,
+    ) -> Tuple[Union[PlotlyObject, pd.DataFrame], Optional[VolcanoPlot]]:
+        pass
+
+    def _nan_check(self) -> None:  # noqa: B027
+        """Raise ValueError for methods that do not tolerate NaNs if there are any."""
+        if not self._works_with_nans and self._dataset.mat.isnull().values.any():
```
JuliaS92, line 81: I know you only moved this, but do we want isnull or isnan here?

alphastats/gui/utils/overview_helper.py
```diff
@@ -21,47 +20,17 @@ def get_intensity_distribution_processed():
...
-                PreprocessingStateKeys.LOG2_TRANSFORMED
-            ]
-        )
-    )
-
     st.markdown("**DataFrame used for analysis** *preview*")
-    st.markdown(text)
 
-    df = get_display_matrix()
+    # TODO why not use the actual matrix here?
```
JuliaS92, line 26: I think this was meant to not show the full column names (proteins).
Actually it would be more important to show all samples, but only 10 proteins and not the other way around.

alphastats/gui/utils/overview_helper.py
```diff
@@ -21,47 +20,17 @@ def get_intensity_distribution_processed():
...
+    df = pd.DataFrame(
+        st.session_state[StateKeys.DATASET].mat.values,
+        index=st.session_state[StateKeys.DATASET].mat.index.to_list(),
+    ).head(10)
 
     st.dataframe(df)
 
-    show_button_download_df(df, file_name="analysis_matrix")
```
JuliaS92, line 63: Why did you remove the download button? I think it's crucial that the processed data can be downloaded.

alphastats/gui/utils/overview_helper.py
```diff
@@ -21,47 +20,17 @@ def get_intensity_distribution_processed():
...
-                PreprocessingStateKeys.IMPUTATION
-            ]
-        )
-        + ", Log2-transformed: "
-        + str(
-            st.session_state[StateKeys.DATASET].preprocessing_info[
-                PreprocessingStateKeys.LOG2_TRANSFORMED
-            ]
-        )
-    )
```
JuliaS92, line 54: I think it would be valid to have a textual representation of the preprocessing here, but I agree with deleting the version as it was. How about we assemble a 'method section' based on the preprocessing parameters?

JuliaS92 (approved):
Just bring the download button back please.

alphastats/gui/utils/overview_helper.py
mschwoer: I found this also highly redundant to the display of preprocessing parameters on the sidebaer. .. a human-readable "Methods" section (that can be pasted into a paper) would be awesome!

alphastats/gui/utils/overview_helper.py
mschwoer: I felt it did not add too much value as only the first 10 rows were downloaded, but now I see that I should have solved that in a bit less destructive way :-D

PR #376, opened 2024-11-18
fix plotting

mschwoer:
one does not simply not set the background color :-D


JuliaS92 (approved):
I did absolutely not expect this. Sorry for the extra work.

PR #386, opened 2024-12-05
Llm uniprot

JuliaS92:
The new formatted uniprot infromation has been integrated as new tool for the LLM.
I tested this a bit and it works nicely.
One question I have is: What would you expect to happen if you already started a chat and then change the selected fields from uniprot? Should the LLM get an update on what it's supposed to know, or is this the users job to decide upfront or reset the chat?

### DIFF
--- a/alphastats/gui/pages/06_LLM.py
+++ b/alphastats/gui/pages/06_LLM.py
@@ -172,6 +172,7 @@ with st.expander("System message", expanded=False):
         disabled=llm_integration_set_for_model,
     )
 
+# TODO: Regenerate initial prompt on reset
 with st.expander("Initial prompt", expanded=True):
     feature_to_repr_map = st.session_state[StateKeys.DATASET]._feature_to_repr_map
     initial_prompt = st.text_area(
--- a/alphastats/gui/pages/06_LLM.py
+++ b/alphastats/gui/pages/06_LLM.py
@@ -173,10 +173,13 @@ with st.expander("System message", expanded=False):
     )
 
 with st.expander("Initial prompt", expanded=True):
+    feature_to_repr_map = st.session_state[StateKeys.DATASET]._feature_to_repr_map
     initial_prompt = st.text_area(
         "",
         value=get_initial_prompt(
-            plot_parameters, upregulated_genes, downregulated_genes
+            plot_parameters,
+            list(map(feature_to_repr_map.get, upregulated_genes)),
+            list(map(feature_to_repr_map.get, downregulated_genes)),
         ),
         height=200,
         disabled=llm_integration_set_for_model,
--- a/alphastats/gui/pages/06_LLM.py
+++ b/alphastats/gui/pages/06_LLM.py
@@ -152,18 +152,20 @@ with c1:
         )
 
 
-st.markdown("##### Select which information from Uniprot to supply to the LLM")
-display_uniprot(
-    regulated_genes_dict, st.session_state[StateKeys.DATASET]._feature_to_repr_map
-)
-
-st.markdown("##### Prompts generated based on analysis input")
-
 model_name = st.session_state[StateKeys.MODEL_NAME]
 llm_integration_set_for_model = (
     st.session_state.get(StateKeys.LLM_INTEGRATION, {}).get(model_name, None)
     is not None
 )
+
+st.markdown("##### Select which information from Uniprot to supply to the LLM")
+display_uniprot(
+    regulated_genes_dict,
+    st.session_state[StateKeys.DATASET]._feature_to_repr_map,
+    disabled=llm_integration_set_for_model,
+)
+
+st.markdown("##### Prompts generated based on analysis input")
 with st.expander("System message", expanded=False):
     system_message = st.text_area(
         "",
--- a/alphastats/llm/llm_functions.py
+++ b/alphastats/llm/llm_functions.py
@@ -41,7 +41,7 @@ def get_annotation_from_uniprot_by_feature_list(
     Args:
         features (list): A list of features to search for.
     Returns:
-        Union[dict, str]: The annotation for the identified feature. The return type can be a dictionaryn or a string depending on the annotation retrieved.
+        Union[dict, str]: The annotation for the identified feature. The return type can be a dictionary or a string depending on the annotation retrieved.
         feature (str): The feature that was used to retrieve the annotation.
     """
     if len(features) == 0:
--- a/alphastats/llm/llm_functions.py
+++ b/alphastats/llm/llm_functions.py
@@ -14,8 +14,8 @@ from alphastats.llm.uniprot_utils import (
 )
 
 
-def get_annotation_from_store_by_repr(
-    feature_repr: str, feature_to_repr_map: dict, annotation_store: dict
+def get_annotation_from_store_by_feature_list(
+    features: list, annotation_store: dict
 ) -> Union[dict, str, None]:
     """
     Retrieve the annotation from the store based on the feature representation.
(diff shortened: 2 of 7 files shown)


tests/llm/test_llm_functions.py
```diff
@@ -93,3 +99,139 @@ def test_assistant_functions():
...
+    get_uniprot_info_for_llm_input(llm_input)
+    mock_get_annotation_from_store.assert_called_with(["id1;id4"], {})
+
+    # protein
+    llm_input = "id1"
+    get_uniprot_info_for_llm_input(llm_input)
+    mock_get_annotation_from_store.assert_called_with(["id1;id4", "id5;id1"], {})
+
+    # not valid
+    llm_input = "id6"
```
mschwoer, line 235: this could be split into ~6 different tests (using `DUMMY_DATASET()` as a fixture)

tests/llm/test_llm_functions.py
```diff
@@ -93,3 +99,139 @@ def test_assistant_functions():
...
+@patch("streamlit.session_state", new_callable=dict)
+def test_get_uniprot_info_for_llm_input(
+    mock_session_state,
+    mock_get_annotation_from_store,
+    mock_get_annotation_from_uniprot,
+    mock_format_uniprot_annotation,
+):
+    """Test that the function retrieves the correct UniProt information for the LLM input."""
+
+    class DUMMY_DATASET:
```
mschwoer, line 178: (nit) `DummyDataset`

tests/llm/test_llm_functions.py
```diff
@@ -93,3 +99,139 @@ def test_assistant_functions():
...
+    mock_format_uniprot_annotation.return_value = ""
+    result = get_uniprot_info_for_llm_input(llm_input)
+    mock_get_annotation_from_store.assert_called_with(["id1;id4"], {})
+    mock_get_annotation_from_uniprot.assert_called_with(["id1;id4"])
+    assert result == "gene2;gene4: "
+
+    mock_get_annotation_from_store.return_value = {"annotation": "annotation1"}
+    # repr that is also gene
+    llm_input = "gene2"
+    get_uniprot_info_for_llm_input(llm_input)
```
mschwoer, line 215: is there a reason why you are not `assert`ing the return value?

tests/llm/test_llm_functions.py
```diff
@@ -93,3 +99,139 @@ def test_assistant_functions():
...
+
+    # different base ids, no longest element
+    features = ["id2", "id3"]
+    get_annotation_from_uniprot_by_feature_list(features)
+    mock_uniprot_annotation.assert_called_with("id2")
+
+    # different base ids, longest element
+    features = ["id2", "id2;id3"]
+    get_annotation_from_uniprot_by_feature_list(features)
+    mock_uniprot_annotation.assert_called_with("id2;id3")
```
mschwoer, line 159: is there a reason why you are not asserting the return value?

alphastats/llm/llm_functions.py
```diff
@@ -1,15 +1,116 @@
...
+    This function searches for a feature in the feature-to-representation map that matches the given feature representation.
+    If a matching feature is found in the annotation store, its corresponding annotation is returned.
+    Args:
+        feature_repr (str): The representation of the feature to search for.
+        feature_to_repr_map (dict): A dictionary mapping features to their representations.
+        annotation_store (dict): A dictionary storing annotations for features.
+    Returns:
+        Union[dict, str, None]: The annotation corresponding to the feature representation if found, otherwise None.
+    """
+    for feature in features:
```
mschwoer, line 31: the other features that come after a match are ignored on purpose?

alphastats/llm/llm_integration.py
```diff
@@ -151,8 +151,9 @@ def _truncate_conversation_history(self, max_tokens: int = 100000):
             The maximum number of tokens to keep in history, by default 100000
         """
         # TODO: avoid important messages being removed (e.g. facts about genes)
+        # TODO: find out how messages can be None type and handle them earlier
```
mschwoer, line 154: this did happen? maybe add some debug code temporarily that prints out the context (probably to `_parse_model_response`:
There we have the whole `response` object. (what we call "message" is actually `response.choices[0].message.content`

alphastats/gui/pages/06_LLM.py
```diff
@@ -172,11 +172,15 @@ def llm_config():
         disabled=llm_integration_set_for_model,
     )
 
+# TODO: Regenerate initial prompt on reset
```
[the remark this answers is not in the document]
JuliaS92: Yes.

alphastats/llm/llm_functions.py
```diff
@@ -1,15 +1,116 @@
...
+
+    Returns:
+        str: The formatted UniProt information for the feature."""
+
+    feature_to_repr_map = st.session_state[StateKeys.DATASET]._feature_to_repr_map
+    gene_to_features_map = st.session_state[StateKeys.DATASET]._gene_to_features_map
+    protein_to_features_map = st.session_state[
+        StateKeys.DATASET
+    ]._protein_to_features_map
+    annotation_store = st.session_state[StateKeys.ANNOTATION_STORE]
```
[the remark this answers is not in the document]
JuliaS92: If it is possible to pass dictionaries to the LLM tools, yes. Otherwise not. This would need to be defined in get_general_assistant_functions. I couldn't find the documentation for allowed types.

alphastats/llm/llm_functions.py
```diff
@@ -1,15 +1,116 @@
...
+    elif llm_input in feature_to_repr_map:
+        features = [llm_input]
+
+    elif llm_input in gene_to_features_map:
+        features = gene_to_features_map[llm_input]
+
+    elif llm_input in protein_to_features_map:
+        features = protein_to_features_map[llm_input]
+    else:
+        raise ValueError(f"{llm_input} not found in dataset.")
```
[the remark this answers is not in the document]
JuliaS92: I think if there is a hallucination in what the LLM tries to retrieve from uniprot that is a valid outcome.

alphastats/llm/llm_functions.py
JuliaS92: yes, the assymption is that only one entry with the repr makes it to the store, and if they do they are most likely identical or the llm couldn't tell them apart either. This is one reason why I'd rather do this controlled once upfront for all regulated features, as I can then pass the unique feature id (new PR).

alphastats/llm/llm_integration.py
JuliaS92: yes, it happened. I will try again. Could be that this was 8B specific

alphastats/llm/llm_functions.py
```diff
@@ -1,15 +1,116 @@
...
+    It retrieves the annotation for the most appropriate feature from UniProt. Most appropriate is defined as 1. the only feature, 2. any feature if they only differ in isoform ids, 3. the feature with the most base ids, as this is most likely to contain a well annotated protein.
+    Args:
+        features (list): A list of features to search for.
+    Returns:
+        Union[dict, str]: The annotation for the identified feature. The return type can be a dictionaryn or a string depending on the annotation retrieved.
+    """
+    if len(features) == 1:
+        feature = features[0]
+    elif len(features) == 0:
+        raise ValueError("No features provided.")
```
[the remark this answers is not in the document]
JuliaS92: done

alphastats/llm/llm_functions.py
JuliaS92: I updates the doc string accordingly

alphastats/llm/llm_functions.py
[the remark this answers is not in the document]
mschwoer: but wouldn't raising here stop the current LLM analysis?

alphastats/gui/pages/06_LLM.py
```diff
@@ -172,11 +172,15 @@ def llm_config():
         disabled=llm_integration_set_for_model,
```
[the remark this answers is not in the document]
JuliaS92: Done, thank you

alphastats/llm/llm_functions.py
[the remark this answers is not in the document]
JuliaS92: Actually it doesn't:
grafik

alphastats/llm/llm_functions.py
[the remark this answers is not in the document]
JuliaS92: I added it to the board.

tests/llm/test_llm_functions.py
JuliaS92: The mock call is the last call in the function and I asserted once in the first test, that its return value (mocked) is the final return value of the tested function. The tests aim at the value the mocked function is called with.

tests/llm/test_llm_functions.py
JuliaS92: Same, as above. I only want to test the call arguments and test the assmebly of the output only once.

alphastats/llm/llm_functions.py
```diff
@@ -1,15 +1,119 @@
...
+        baseid_sets = []
+        for feature in features:
+            baseid_sets.append(
+                set([identifier.split("-")[0] for identifier in feature.split(";")])
+            )
+        if len(set.union(*baseid_sets)) == len(set.intersection(*baseid_sets)):
+            feature = features[0]
+        else:
+            feature = features[baseid_sets.index(max(baseid_sets, key=len))]
+    annotation = get_annotations_for_feature(feature)
```
boopthesnoot, line 61: maybe we could always return a dict, even when there is a failure?

alphastats/llm/llm_functions.py
```diff
@@ -1,15 +1,119 @@
...
+def get_annotation_from_uniprot_by_feature_list(
+    features: list,
+) -> Tuple[Union[dict, str], str]:
+    """
+    Retrieve annotation from UniProt based on a list of features.
+    It retrieves the annotation for the most appropriate feature from UniProt. Most appropriate is defined as 1. the only feature, 2. any feature if they only differ in isoform ids, 3. the feature with the most base ids, as this is most likely to contain a well annotated protein.
+    Args:
+        features (list): A list of features to search for.
+    Returns:
+        Union[dict, str]: The annotation for the identified feature. The return type can be a dictionaryn or a string depending on the annotation retrieved.
```
boopthesnoot, line 44: fixed typo
```suggestion
        Union[dict, str]: The annotation for the identified feature. The return type can be a dictionary or a string depending on the annotation retrieved.
```

boopthesnoot (approved):

alphastats/llm/llm_functions.py
JuliaS92: I actually prefer, that along the way strings are returned and passed on, as this might inform the LLM as to why there was no information for a specific entry, rather than just providing nothing. We could then include in the system prompt, that certain annotations should lead to specific outcomes, like rating a protein more lowly in it's weight on the biological interpretation.

alphastats/llm/llm_integration.py
JuliaS92: I left this as a todo.

PR #387, opened 2024-12-11
Add download button to get chatlog.

JuliaS92:
First vertical slice of chat saving as simple txt file. I excluded tool replies, as they ususally duplicate in the form of artifacts (correct me if I'm wrong).

### DIFF
--- a/alphastats/gui/pages/06_LLM.py
+++ b/alphastats/gui/pages/06_LLM.py
@@ -263,7 +263,14 @@ def llm_chat(llm_integration: LLMIntegration, show_all: bool = False):
     if prompt := st.chat_input("Say something"):
         with st.spinner("Processing prompt..."):
             llm_integration.chat_completion(prompt)
-        st.rerun(scope="fragment")
+            st.rerun(scope="fragment")
+
+    st.download_button(
+        "Download chat log",
+        llm_integration.get_chat_log_txt(),
+        f"chat_log_{model_name}.txt",
+        "text/plain",
+    )
 
 
 show_all = st.checkbox(
--- a/alphastats/gui/pages/06_LLM.py
+++ b/alphastats/gui/pages/06_LLM.py
@@ -263,7 +263,7 @@ def llm_chat(llm_integration: LLMIntegration, show_all: bool = False):
     if prompt := st.chat_input("Say something"):
         with st.spinner("Processing prompt..."):
             llm_integration.chat_completion(prompt)
-            st.rerun(scope="fragment")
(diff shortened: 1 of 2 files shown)


alphastats/gui/pages/06_LLM.py
```diff
@@ -263,7 +263,14 @@ def llm_chat(llm_integration: LLMIntegration, show_all: bool = False):
     if prompt := st.chat_input("Say something"):
         with st.spinner("Processing prompt..."):
             llm_integration.chat_completion(prompt)
-        st.rerun(scope="fragment")
+            st.rerun(scope="fragment")
```
mschwoer, line 266: why change the indent here?

alphastats/llm/llm_integration.py
```diff
@@ -289,6 +289,21 @@ def get_print_view(self, show_all=False) -> List[Dict[str, Any]]:
...
+        messages = self.get_print_view(show_all=True)
+        chatlog = ""
+        for message in messages:
+            if message["role"] == "tool":
+                continue
+            chatlog += f"{message['role'].capitalize()}: {message['content']}\n"
+            if len(message["artifacts"]) > 0:
+                chatlog += "-----\n"
+            for artifact in message["artifacts"]:
+                chatlog += f"Artifact: {artifact}\n"
```
mschwoer, line 303: would be cool if e.g. plotly plots could be actually displayed by copying this string (might need some escaping)

mschwoer (approved):
LGTM

alphastats/gui/pages/06_LLM.py
JuliaS92: my bad

alphastats/llm/llm_integration.py
JuliaS92: This is possible with some alterations to the outputted string as it is now. After trying I identified these points that need changing, but this list is likely incomplete and automating it could be detrimental (as I think ... also gets introduced to the data arrays if they become too long):
- remove any lines with ..., which includes hovertemplates and the layout template (therefore becomes a bit uglier)
- add quotation marks to strings in axis category arrays

Then one can reproduce the figure with this code, copying the curly bracket from the artifact output.
```
import plotly.graph_objects as go
from numpy import array
go.Figure(**{
    'data': ...,
    'layout': ...
})
```

This could be mitigated by exporting the figures to html, but then I would suggest one file per figure.

mschwoer: hm, seems quite brittle indeed .. then just leave it for now

PR #388, opened 2024-12-16
Validation and refactoring of differential expression analysis

JuliaS92:
This is a base branch for a major effort to refactor the differential expression analysis.

Currently this covers a skeleton for a new 3-tier class system for all differential expression analysis:
1. DiffernetialExpressionAnalysis: enforces unified input and output
2. DifferentialExpressionAnalysis_TwoGroups: validation of groups shared, but unique funcitonality
3. Concrete statistical test, e.g. DifferentialExpressionAnalysis_TTest

If you review at this stage, please check in particular if
- validation and passing of parameters is sufficiently robust and not overengineered (already a lot beeter after Magnus's review)
- Check the TODOs added in other files to see how I envision to integrate this with the rest of the tool
- any of the branches below are not supported by the code yet

Checklist to do before branching off:
- [x] Write tests
- [x] Talk to an AI about this
- [x] Check if this is compatible with all statistical tests that are supposed to run in the end
 - The only exception is the multicova, since this needs to be refactored internally first.
- [x] Check if this easily interfaces with UI and plotting

Branches that can then be based on this:
- [ ] Run and store in Dataset
- [ ] Recreate Volcano plot from result, significance and dataset
- [ ] UI integration via gui/utils/analysis
- [ ] Depracate old functionality (DifferentialExpressionAnalysis without plot), keep thin wrapper for multicova until refactoring.
- [ ] Individual statistical test classes, including actual test of the static functions (not included for the t-test here on purpose, will be a separate branch).

### DIFF
--- a/alphastats/dataset/dataset.py
+++ b/alphastats/dataset/dataset.py
@@ -90,6 +90,8 @@ class DataSet:
 
         # self.evidence_df: pd.DataFrame = loader.evidence_df  # TODO unused
 
+        # TODO: Add a store for dea results here
+
         self._dataset_factory = DataSetFactory(
             rawinput=self.rawinput,
             intensity_column=self._intensity_column,
@@ -269,6 +271,7 @@ class DataSet:
             preprocessing_info=self.preprocessing_info,
         )
 
+    # TODO: Add function get_differential_expression_analysis() which will handle the dea store and run diff_expression_analysis() if necessary
     def diff_expression_analysis(
         self,
         group1: Union[str, list],
@@ -279,6 +282,7 @@ class DataSet:
         fdr: float = 0.05,
     ) -> pd.DataFrame:
         """A wrapper for the Statistics.diff_expression_analysis(), see documentation there."""
+        # TODO: This method is the one which will be called if a dea result is not yet in the store.
         return self._get_statistics().diff_expression_analysis(
             group1,
             group2,
--- a/alphastats/tl/differential_expression_analysis.py
+++ b/alphastats/tl/differential_expression_analysis.py
@@ -282,7 +282,28 @@ class DifferentialExpressionAnalysisTwoGroups(DifferentialExpressionAnalysis, AB
 
 
 class DifferentialExpressionAnalysisTTest(DifferentialExpressionAnalysisTwoGroups):
-    """This class implements the t-test differential expression analysis."""
+    """This class implements the t-test differential expression analysis.
+
+    Examples:
+    >>> mat = pd.DataFrame({
+    ...     'sample1': [1, 2, 3],
+    ...     'sample2': [4, 5, 6],
+    ...     'sample3': [7, 8, 9],
+    ...     'sample4': [10, 11, 12]
+    ... }, index=['gene1', 'gene2', 'gene3']).T
+    >>> dea = DifferentialExpressionAnalysisTTest(mat)
+    >>> result = dea.perform(
+    ...     group1_samples = ['sample1', 'sample2'],
+    ...     group2_samples = ['sample3', 'sample4'],
+    ...     is_log2_transformed = True,
+    ...     test_type = "independent",
+    ...     fdr_method = "fdr_bh",
+    ... )
+    >>> result.columns.tolist()
+    ['p-value', 'log2(fold change)', 'q-value']
+    >>> result.index.tolist()
+    ['gene1', 'gene2', 'gene3']
+    """
 
     @staticmethod
     def _allowed_parameters() -> List[str]:
--- a/alphastats/tl/differential_expression_analysis.py
+++ b/alphastats/tl/differential_expression_analysis.py
@@ -7,7 +7,6 @@ import scipy
 from statsmodels.stats.multitest import multipletests
 
 from alphastats.dataset.keys import Cols, ConstantsClass
-from alphastats.dataset.preprocessing import PreprocessingStateKeys
 from alphastats.statistics.statistic_utils import calculate_foldchange
 
 
@@ -20,6 +19,7 @@ class DeaColumns(ConstantsClass):
 
 class DeaParameters(ConstantsClass):
     METADATA = "metadata"
+    ISLOG2TRANSFORMED = "is_log2_transformed"
 
     GROUP1 = "group1"
     GROUP2 = "group2"
@@ -318,7 +318,7 @@ class DifferentialExpressionAnalysisTTest(DifferentialExpressionAnalysisTwoGroup
             DeaParameters.GROUP2,
             DeaParameters.GROUPING_COLUMN,
             DeaParameters.METADATA,
-            PreprocessingStateKeys.LOG2_TRANSFORMED,
+            DeaParameters.ISLOG2TRANSFORMED,
         ]
 
     def _extend_validation(
@@ -345,7 +345,17 @@ class DifferentialExpressionAnalysisTTest(DifferentialExpressionAnalysisTwoGroup
         if fdr_method not in ["fdr_bh", "bonferroni"]:
             raise ValueError("fdr_method must be one of 'fdr_bh', 'bonferroni'.")
 
-    def _perform(self, **kwargs) -> pd.DataFrame:
+    def _perform(
+        self,
+        is_log2_transformed: bool,
+        test_type: str,
+        fdr_method: str,
+        group1: Union[List, str],
+        group2: Union[List, str],
+        grouping_column: Union[str, None] = None,
+        metadata: Union[pd.DataFrame, None] = None,
+        **kwargs,
+    ) -> pd.DataFrame:
         """Runs the t-test analysis and returns the result.
         Wrapper to staistical method with actual method parameters and implementation.
 
@@ -353,19 +363,19 @@ class DifferentialExpressionAnalysisTTest(DifferentialExpressionAnalysisTwoGroup
         pd.DataFrame: The result of the analysis.
         """
         group1_samples, group2_samples = self._get_group_members(
-            group1=kwargs[DeaParameters.GROUP1],
-            group2=kwargs[DeaParameters.GROUP2],
-            grouping_column=kwargs.get(DeaParameters.GROUPING_COLUMN, None),
-            metadata=kwargs.get(DeaParameters.METADATA, None),
+            group1=group1,
+            group2=group2,
+            grouping_column=grouping_column,
+            metadata=metadata,
         )
 
         result = self._run_statistical_test(
             mat=self.mat,
             group1_samples=group1_samples,
             group2_samples=group2_samples,
-            is_log2_transformed=kwargs[PreprocessingStateKeys.LOG2_TRANSFORMED],
-            test_type=kwargs[DeaParameters.TEST_TYPE],
-            fdr_method=kwargs[DeaParameters.FDR_METHOD],
+            is_log2_transformed=is_log2_transformed,
+            test_type=test_type,
+            fdr_method=fdr_method,
         )
         return result
 
--- a/alphastats/tl/differential_expression_analysis.py
+++ b/alphastats/tl/differential_expression_analysis.py
@@ -225,11 +225,11 @@ class DifferentialExpressionAnalysisTTest(DifferentialExpressionAnalysisTwoGroup
         """
         super()._extend_validation(input_data, parameters)
         if parameters["test_fun"] not in [
-            scipy.stats.ttest_ind,
-            scipy.stats.ttest_rel,
+            "independent",
+            "paired",
         ]:
             raise ValueError(
-                "test_fun must be either scipy.stats.ttest_ind or scipy.stats.ttest_rel for t-test."
+                "test_fun must be either 'independent' for scipy.stats.ttest_ind or 'paired' for scipy.stats.ttest_rel."
             )
         if parameters["fdr_method"] not in ["bh", "by"]:
             raise ValueError("fdr_method must be one of 'bh', 'by'.")
@@ -258,7 +258,7 @@ class DifferentialExpressionAnalysisTTest(DifferentialExpressionAnalysisTwoGroup
         group1: list,
         group2: list,
         is_log2_transformed: bool,
-        test_fun: callable,
+        test_fun: str,
         fdr_method: str,
     ) -> pd.DataFrame:
         """Runs the t-test analysis and returns the result.
@@ -268,13 +268,18 @@ class DifferentialExpressionAnalysisTTest(DifferentialExpressionAnalysisTwoGroup
         group1 (list): The samples for group 1.
         group2 (list): The samples for group 2.
         is_log2_transformed (bool): Whether the data is log2 transformed.
-        test_fun (callable): The test function to use, scipy.stats.ttest_ind or scipy.stats.ttest_rel.
+        test_fun (str): The test function to use, independent for scipy.stats.ttest_ind or paired for scipy.stats.ttest_rel.
         fdr_method (str): The FDR method to use, 'bh' or 'by'.
 
         Returns:
         pd.DataFrame: The result of the analysis.
         """
         mat_transpose = input_data.loc[group1 + group2, :].transpose()
+
+        test_fun = {
+            "independent": scipy.stats.ttest_ind,
+            "paired": scipy.stats.ttest_rel,
+        }[test_fun]
 
         if not is_log2_transformed:
             mat_transpose = mat_transpose.transform(np.log2)
--- a/alphastats/tl/differential_expression_analysis.py
+++ b/alphastats/tl/differential_expression_analysis.py
@@ -63,36 +63,34 @@ class DifferentialExpressionAnalysis(ABC):
     """
 
     def __init__(self, mat: pd.DataFrame) -> None:
-        """Constructor for the DifferentialExpressionAnalysis class. Validates input and parameters.
+        """Constructor for the DifferentialExpressionAnalysis class. sets up mat and results.
 
         Parameters:
         mat (pd.DataFrame): The input data for the analysis. This should be a DataFrame with the samples as rows and the features as columns.
         """
+        if mat.empty:
+            raise ValueError(
+                "Input matrix to differential expression analysis is empty."
+            )
         self.mat = mat
         self.result: pd.DataFrame = None
 
-    def _validate_input(self, parameters: dict) -> None:
+    def _validate_input(self, **kwargs) -> None:
         """Abstract method to validate the input and parameters. This should raise an exception if the input or parameters are invalid
 
         This function here checks for all parameters required for analysis regardless of the specific method, namely log2_transformed and metadata.
 
         Parameters:
-        mat (pd.DataFrame): The input data for the analysis.
-        parameters (dict): The parameters for the analysis.
+        **kwargs (dict): The parameters for the analysis. The keys need to be defined within the allowed_parameters method.
         """
-        if self.mat is None:
-            raise ValueError("No input data was provided.")
-        if parameters is None:
-            raise ValueError("No parameters were provided.")
-
         try:
-            self._extend_validation(**parameters)
+            self._extend_validation(**kwargs)
         except TypeError as err:
             raise TypeError(
                 f"{str(err)}. Accepted keyword arguments to perform are {', '.join(self._allowed_parameters())}."
             ) from err
 
-        for parameter in parameters:
+        for parameter in kwargs:
             if parameter not in self._allowed_parameters():
                 raise TypeError(
                     f"Parameter {parameter} should not be provided for this analysis. Accepted keyword arguments to perform are {', '.join(self._allowed_parameters())}."
@@ -106,6 +104,8 @@ class DifferentialExpressionAnalysis(ABC):
 
     @abstractmethod
     def _extend_validation(self, **kwargs) -> None:
+        """Abstract method to extend the validation of parameters for the specific method. This should raise an exception if the input or parameters are invalid.
+        It should have all parameters required for the method as keyword arguments and validate that parameters are compatible with mat."""
         pass
 
     def perform(self, **kwargs) -> Tuple[str, pd.DataFrame]:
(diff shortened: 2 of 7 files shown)


tests/tl/test_differential_expression_analysis.py
```diff
@@ -0,0 +1,412 @@
...
+            index=["gene1"],
+        )
+
+
+def test_dea_no_abstractmethods():
+    with pytest.raises(TypeError):
+        DifferentialExpressionAnalysis(None)
+
+
+def test_dea_perform_success():
```
mschwoer, line 46: docstrings would help here, also using `match` for the individual `Errors` would increase readability and robustness (there are multiple different `Value` and `KeyErrors`)

tests/tl/test_differential_expression_analysis.py
```diff
@@ -0,0 +1,412 @@
...
+    input_data = pd.DataFrame()
+    dea = TestableDifferentialExpressionAnalysis(input_data)
+    dea.perform(metadata=pd.DataFrame())
+    assert dea.result.equals(
+        pd.DataFrame(
+            [[0.055, 2, 0.009]],
+            columns=[DeaColumns.PVALUE, DeaColumns.LOG2FC, DeaColumns.QVALUE],
+            index=["gene1"],
+        )
```
mschwoer, line 55: I was first puzzled why there is an result for empty input ..
sugestions: 
- make this df a global variable and use it here (and everywhere else) and in `TestableDifferentialExpressionAnalysis` or
- patch `_statistical_test_fun` and `assert_called_once_with` (preferred, at least once)

tests/tl/test_differential_expression_analysis.py
```diff
@@ -0,0 +1,412 @@
...
+
+def test_dea_get_dict_key_static():
+    assert (
+        DifferentialExpressionAnalysis.get_dict_key(
+            {DeaParameters.METADATA: pd.DataFrame()}
+        )
+        == """{'metadata': Empty DataFrame
+Columns: []
+Index: []}"""
+    )
```
mschwoer, line 186: we could drop the whole `get_dict_key` for now until we need it

alphastats/tl/differential_expression_analysis.py
```diff
@@ -0,0 +1,345 @@
...
+
+        Returns:
+        pd.DataFrame: A DataFrame with a single binary column.
+        """
+        significance = pd.DataFrame(index=result.index)
+        significance[DeaColumns.SIGNIFICANT] = result[DeaColumns.QVALUE] < qvalue_cutoff
+        return significance
+
+
+class DifferentialExpressionAnalysisTwoGroups(DifferentialExpressionAnalysis):
```
mschwoer, line 184: `class DifferentialExpressionAnalysisTwoGroups(DifferentialExpressionAnalysis, ABC):`

tests/tl/test_differential_expression_analysis.py
```diff
@@ -0,0 +1,412 @@
...
+    ).T
+
+
+def test_dea_two_groups_no_abstractmethods():
+    with pytest.raises(TypeError):
+        DifferentialExpressionAnalysisTwoGroups(None)
+
+
+def test_dea_two_groups_perform_success():
+    input_data = TestableDifferentialExpressionAnalysisTwoGroups.valid_data_input
```
mschwoer, line 241: could we make `valid_data_input` a global constant and use it here and for defining `TestableDifferentialExpressionAnalysisTwoGroups`? Seeing `TestableDifferentialExpressionAnalysisTwoGroups` and an "real" instance at the same place is a bit misleadding.

tests/tl/test_differential_expression_analysis.py
```diff
@@ -0,0 +1,412 @@
...
+    dea = DifferentialExpressionAnalysisTTest(input_data)
+    dea.perform(
+        **TestableDifferentialExpressionAnalysisTwoGroups.valid_parameter_input,
+        **{
+            DeaParameters.TEST_FUN: "independent",
+            DeaParameters.FDR_METHOD: "fdr_bh",
+            PreprocessingStateKeys.LOG2_TRANSFORMED: True,
+        },
+    )
+    assert dea.result.shape == (3, 3)
```
mschwoer, line 362: At least for one set of parameters, I would test the actual results.
`pd.testing.assert_frame_equals` could come handy here

alphastats/tl/differential_expression_analysis.py
```diff
@@ -0,0 +1,345 @@
...
+        pd.DataFrame: The result of the analysis.
+        """
+        group1, group2 = self._get_group_members(kwargs)
+        result = self._statistical_test_fun(
+            input_data=self.input_data,
+            group1=group1,
+            group2=group2,
+            is_log2_transformed=kwargs[PreprocessingStateKeys.LOG2_TRANSFORMED],
+            test_fun=kwargs[DeaParameters.TEST_FUN],
+            fdr_method=kwargs[DeaParameters.FDR_METHOD],
```
mschwoer, line 286: I would strongly prefer explicit parameters here

mschwoer (commented):
Look quite good already!

alphastats/dataset/dataset.py
```diff
@@ -90,6 +90,8 @@ def __init__(
 
         # self.evidence_df: pd.DataFrame = loader.evidence_df  # TODO unused
 
+        # TODO: Add a store for dea results here
```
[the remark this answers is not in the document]
JuliaS92: For starters I would just go with a dictionary, but zes I will keept this in mind when I start the branch for this.

alphastats/tl/differential_expression_analysis.py
```diff
@@ -0,0 +1,345 @@
...
+
+    attributes:
+    - input_data (pd.DataFrame): The input data for the analysis.
+    - result (pd.DataFrame): The result of the analysis.
+
+    abstract methods:
+    - _allowed_parameters: that returns a list of allowed parameters for the analysis, static method
+    - _extend_validation: extends the validation of parameters for the specific method
+    - _run_statistical_test: wrapper runs the statistical test from kwargs
+    - _statistical_test_fun: that runs the statistical test, static method. The wrapper and the actual method are separated to allow for easier testing and to ensure that all parameters are defined with types and defaults.
```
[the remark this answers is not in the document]
JuliaS92: I thought it would be nice to get an overview of the abstract mathods that need to be implemented by children. I'll move it down if this is the optionsal part.

alphastats/tl/differential_expression_analysis.py
```diff
@@ -0,0 +1,345 @@
...
+
+    public methods:
+    - perform: performs the analysis and stores the result. This fixes the worflow of input validation, running the test and validating the output.
+    - get_dict_key: generates a unique key for the result dictionary, static method
+    - get_significance: returns the significant features based on the q-value cutoff, static method
+
+    Intended usage:
```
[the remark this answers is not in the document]
JuliaS92: I can add a doctest to the TTest class. Here I would need to actually implement the methods for a working doctest.

alphastats/tl/differential_expression_analysis.py
```diff
@@ -0,0 +1,345 @@
...
+        implement t-test specific methods
+    dea = DifferentialExpressionAnalysisTTest(DataSet.mat)
+    settings = {'group1': ['A', 'B'], 'group2': ['C', 'D'], 'test_fun': 'independent', 'fdr_method': 'fdr_bh'}
+    result = dea.perform(**settings) # run once
+    cached_results[dea.get_dict_key(settings)] = result
+    significance = dea.get_significance(cached_results[dea.get_dict_key(settings)], 0.05) # run multiple times
+    volcano_plot(cached_results[dea.get_dict_key(settings)], significance) # visualize
+    """
+
+    def __init__(self, input_data: pd.DataFrame) -> None:
```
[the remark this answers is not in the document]
JuliaS92: Yes this will be mat (see intended usage above ;) )

alphastats/tl/differential_expression_analysis.py
```diff
@@ -0,0 +1,345 @@
...
+        self._extend_validation(parameters)
+
+        for parameter in parameters:
+            if parameter not in self._allowed_parameters():
+                raise ValueError(
+                    f"Parameter {parameter} should not be provided for this analysis."
+                )
+
+    @staticmethod
+    @abstractmethod
```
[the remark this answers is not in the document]
JuliaS92: This is just to say, that the implementation should be a static method.

alphastats/tl/differential_expression_analysis.py
```diff
@@ -0,0 +1,345 @@
...
+        input_data (pd.DataFrame): The input data for the analysis.
+        **kwargs: The parameters for the analysis. The keys need to be defined within the allowed_parameters method.
+
+        Returns:
+        pd.DataFrame: The result of the analysis.
+        """
+        pass
+
+    @staticmethod
+    def get_significance(result: pd.DataFrame, qvalue_cutoff: float) -> pd.DataFrame:
```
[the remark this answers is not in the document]
JuliaS92: For the store I currently plan to not actually store the dea object, but only the result dataframe. Hence a static method.

alphastats/tl/differential_expression_analysis.py
```diff
@@ -0,0 +1,345 @@
...
+        p_values = mat_transpose.apply(
+            lambda row: test_fun(
+                row[group1].values.flatten(),
+                row[group2].values.flatten(),
+                nan_policy="omit",
+            )[1],
+            axis=1,
+        )
+
+        result = pd.DataFrame(index=input_data.columns)
```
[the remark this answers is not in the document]
JuliaS92: The mat is with proteins in columns (very ML centric), for plotting and also table display of these results they should be in rows (user centric). So i need to flip here, but this is actuallya lso a bug :)

alphastats/tl/differential_expression_analysis.py
```diff
@@ -0,0 +1,345 @@
...
+                raise KeyError(f"Sample {index} is missing from the input data.")
+
+    @staticmethod
+    def _get_group_members(parameters) -> Tuple[list, list]:
+        """Returns the group columns based on the parameters."""
+        if DeaParameters.GROUPING_COLUMN not in parameters:
+            group1 = parameters[DeaParameters.GROUP1]
+            group2 = parameters[DeaParameters.GROUP2]
+        else:
+            grouping_column = parameters[DeaParameters.METADATA][
```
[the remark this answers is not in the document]
JuliaS92: _values

alphastats/tl/differential_expression_analysis.py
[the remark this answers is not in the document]
boopthesnoot: +1, I like it being visible in the class definition. Although keeping it up to date would mean double bookkeeping

alphastats/tl/differential_expression_analysis.py
```diff
@@ -0,0 +1,345 @@
...
+        group1 (list): The samples for group 1.
+        group2 (list): The samples for group 2.
+        is_log2_transformed (bool): Whether the data is log2 transformed.
+        test_fun (str): The test function to use, independent for scipy.stats.ttest_ind or paired for scipy.stats.ttest_rel.
+        fdr_method (str): The FDR method to use, 'fdr_bh' or 'bonferroni'.
+
+        Returns:
+        pd.DataFrame: The result of the analysis.
+        """
+        mat_transpose = input_data.loc[group1 + group2, :].transpose()
```
boopthesnoot, line 312: just a thought: would it be possible to precompute indices for each group after the preprocessing? otherwise this could be slow, esp for larger dataframes

alphastats/tl/differential_expression_analysis.py
```diff
@@ -0,0 +1,345 @@
...
+
+        test_fun = {
+            "independent": scipy.stats.ttest_ind,
+            "paired": scipy.stats.ttest_rel,
+        }[test_fun]
+
+        if not is_log2_transformed:
+            mat_transpose = mat_transpose.transform(np.log2)
```
boopthesnoot, line 320: do we trust that at this point we won't have any zero values in `mat_transpose`?

alphastats/tl/differential_expression_analysis.py
```diff
@@ -0,0 +1,345 @@
...
+            mat_transpose = mat_transpose.replace([np.inf, -np.inf], np.nan)
+
+        mat_transpose = mat_transpose.dropna(how="all")
+
+        # TODO: return not only the p-value, but also the t-statistic
+        p_values = mat_transpose.apply(
```
boopthesnoot, line 326: do we want to fail if a minimum sample size is not reached? otherwise our assumptions might be broken

this can be non obvious to the user, because we're dropping NAs under the hood without telling how many samples we have left

alphastats/tl/differential_expression_analysis.py
```diff
@@ -0,0 +1,345 @@
...
+        This function here checks for all parameters required for analysis regardless of the specific method, namely log2_transformed and metadata.
+
+        Parameters:
+        input_data (pd.DataFrame): The input data for the analysis.
+        parameters (dict): The parameters for the analysis.
+        """
+        if self.input_data is None:
+            raise ValueError("No input data was provided.")
+        if parameters is None:
+            raise ValueError("No parameters were provided.")
```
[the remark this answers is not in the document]
boopthesnoot: maybe the idea was to check if those aren't empty?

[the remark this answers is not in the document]
JuliaS92: I moved the data check to the init and all parameters should now run against a type error if they are missing, or raise a custom error if they are not allowed.

alphastats/tl/differential_expression_analysis.py
JuliaS92: That's why we have the flag for the log_2_transformation. If it wasn't tranformed before the transformation will be done on the fly and remove 0s.

alphastats/tl/differential_expression_analysis.py
JuliaS92: I'm dropping all nan proteins, not samples.
The t-test internally omits nans. We could include an output column for the test with the number of non-na samples used.

alphastats/tl/differential_expression_analysis.py
JuliaS92: Done throughout. Wherever I can predict the arguments they are now explicit. This removed many tests and validations as well.

alphastats/tl/differential_expression_analysis.py
```diff
@@ -0,0 +1,345 @@
...
+        is_log2_transformed (bool): Whether the data is log2 transformed.
+        test_fun (str): The test function to use, independent for scipy.stats.ttest_ind or paired for scipy.stats.ttest_rel.
+        fdr_method (str): The FDR method to use, 'fdr_bh' or 'bonferroni'.
+
+        Returns:
+        pd.DataFrame: The result of the analysis.
+        """
+        mat_transpose = input_data.loc[group1 + group2, :].transpose()
+
+        test_fun = {
```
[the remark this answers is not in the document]
JuliaS92: I would like to keep this as is, for (over)planning for the future, when this might become a public method

alphastats/tl/differential_expression_analysis.py
boopthesnoot: i didn't quite follow where do we remove 0s

JuliaS92: the log2 transformation return np.nan for 0s
I added a non-zero and a all-nan line in the tests to illustrate the bahaviour.

alphastats/tl/differential_expression_analysis.py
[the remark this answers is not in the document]
mschwoer: the double bookkeeping is my concern here.. and in this case, it's quite easy to find out what needs to be implemented: the `ABC` decorator takes care of that ;-)

alphastats/tl/differential_expression_analysis.py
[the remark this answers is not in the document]
mschwoer: I was more referring to the doctest style, e.g. inline by `>` etc

alphastats/tl/differential_expression_analysis.py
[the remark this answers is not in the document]
mschwoer: `static` + `abtract` is fine.. but `abstract` + `return []` is unusual

alphastats/tl/differential_expression_analysis.py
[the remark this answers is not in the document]
mschwoer: aye, then maybe flip already input data? or add this as a comment here

alphastats/tl/differential_expression_analysis.py
mschwoer: maybe add a comment at the respective line that this implicitly removes 0s

alphastats/tl/differential_expression_analysis.py
mschwoer: on the "not obvious to user": we should think of a lightweight mechanism to warn the user (both in console and in frontend) .. e.g. use the `warnings` module, and have the frontend catch it .. => https://github.com/orgs/MannLabs/projects/12/views/1?pane=issue&itemId=91601317

JuliaS92: https://github.com/orgs/MannLabs/projects/12?pane=issue&itemId=91616519

Ticket, added UserWarnings.

alphastats/tl/differential_expression_analysis.py
```diff
@@ -0,0 +1,431 @@
...
+        Parameters:
+        **kwargs (dict): The parameters for the analysis. The keys need to be defined within the allowed_parameters method.
+        """
+        try:
+            self._extend_validation(**kwargs)
+        except TypeError as err:
+            raise TypeError(
+                f"{str(err)}. Accepted keyword arguments to perform are {', '.join(self._allowed_parameters())}."
+            ) from err
+
```
mschwoer, line 96: is it still needed to catch a TypeError here?

alphastats/tl/differential_expression_analysis.py
```diff
@@ -0,0 +1,431 @@
...
+            if parameter not in self._allowed_parameters():
+                raise TypeError(
+                    f"Parameter {parameter} should not be provided for this analysis. Accepted keyword arguments to perform are {', '.join(self._allowed_parameters())}."
+                )
+
+    def _allowed_parameters(self) -> List[str]:
+        """Method returning a list of allowed parameters for the analysis to avoid calling tests with additional parameters."""
+        perform_signature = inspect.signature(self._perform)
+        parameters = list(perform_signature.parameters.keys())
+        return [parameter for parameter in parameters if parameter != "kwargs"]
```
mschwoer, line 107: can be removed, python does that for you :-)

alphastats/tl/differential_expression_analysis.py
```diff
@@ -0,0 +1,431 @@
...
+        Parameters:
+        group1 (Union[List, str]): The first group.
+        group2 (Union[List, str]): The second group.
+        grouping_column (Union[str, None]): The column in the metadata to group by.
+        metadata (Union[pd.DataFrame, None]): The metadata DataFrame.
+
+        Returns:
+        Tuple[list, list]: The samples for group 1 and group 2."""
+        if grouping_column is None:
+            if not isinstance(group1, list) or not isinstance(group2, list):
```
mschwoer, line 241: the three checks on `isinstance` could be moved to `_extend_validation` (but at the cost that the `if grouping_column is None: . .else` logic would be duplicated ..)

alphastats/tl/differential_expression_analysis.py
```diff
@@ -0,0 +1,431 @@
...
+        ]:
+            raise ValueError(
+                f"test_type must be either '{DeaTestTypes.INDEPENDENT}' for scipy.stats.ttest_ind or '{DeaTestTypes.PAIRED}' for scipy.stats.ttest_rel."
+            )
+        if fdr_method not in ["fdr_bh", "bonferroni"]:
+            raise ValueError("fdr_method must be one of 'fdr_bh', 'bonferroni'.")
+
+    def _perform(
+        self,
+        is_log2_transformed: bool,
```
mschwoer, line 325: `is_log2_transformed` is an attribute of the `input_data` and should be set together with it .. otherwise, this added freedom allows for wrong calculations

tests/tl/test_differential_expression_analysis.py
```diff
@@ -0,0 +1,349 @@
...
+            [0.01570651089572518, -1.7237294884856105, 0.047119532687175544],
+            [0.1556022268654943, -1.1111962106682238, 0.23340334029824145],
+            [0.29794294575639113, -0.5, 0.29794294575639113],
+        ],
+        columns=[DeaColumns.PVALUE, DeaColumns.LOG2FC, DeaColumns.QVALUE],
+        index=["gene1", "gene2", "gene3"],
+    )
+    dea = DifferentialExpressionAnalysisTTest(valid_data_input_two_groups)
+    with pytest.warns(UserWarning, match="log2 transformation"):
+        result = dea._perform(
```
mschwoer, line 308: (nit) would prefer testing the public method here

tests/tl/test_differential_expression_analysis.py
```diff
@@ -0,0 +1,349 @@
...
+            **valid_parameter_input_two_groups,
+            **{
+                DeaParameters.TEST_TYPE: DeaTestTypes.INDEPENDENT,
+                DeaParameters.FDR_METHOD: "fdr_bh",
+                DeaParameters.ISLOG2TRANSFORMED: False,
+            },
```
mschwoer, line 314: could we pass those parameters explicitly?

alphastats/tl/differential_expression_analysis.py
JuliaS92: Nope, indeed not.

alphastats/tl/differential_expression_analysis.py
JuliaS92: Not if I pass kwargs to _perform though, right? Which I need to do to match the signature of the abstract method? The idea was that by adding this here I enforce kwargs to be empty without relying on every implementation of _perform to contain that check.

alphastats/tl/differential_expression_analysis.py
JuliaS92: The deduplication of the logic is my I moved it here, but wrote a test to make sure _get_group_members is always called by _extend_validation.

mschwoer (approved):
LGTM! this should be a solid basis for all the upcoming analysis!

PR #389, opened 2024-12-20
dynamically choosing proteins from volcano plot

boopthesnoot:
Added functionality to choose which proteins to look up in uniprot and subsequently analyze with an LLM

### DIFF
--- a/alphastats/gui/pages/06_LLM.py
+++ b/alphastats/gui/pages/06_LLM.py
@@ -11,9 +11,9 @@ from alphastats.gui.utils.analysis_helper import (
     gather_uniprot_data,
 )
 from alphastats.gui.utils.llm_helper import (
-    create_protein_editor,
     display_uniprot,
     llm_connection_test,
+    protein_selector,
     set_api_key,
 )
 from alphastats.gui.utils.ui_helper import (
@@ -136,14 +136,14 @@ with c1:
     c11, c12 = st.columns((1, 1), gap="medium")
 
     with c11:
-        selected_upregulated_genes = create_protein_editor(
+        selected_upregulated_genes = protein_selector(
             upregulated_genes_df,
             "Upregulated Proteins",
             st.session_state[StateKeys.DATASET]._feature_to_repr_map,
         )
 
     with c12:
-        selected_downregulated_genes = create_protein_editor(
+        selected_downregulated_genes = protein_selector(
             downregulated_genes_df,
             "Downregulated Proteins",
             st.session_state[StateKeys.DATASET]._feature_to_repr_map,
--- a/alphastats/gui/pages/06_LLM.py
+++ b/alphastats/gui/pages/06_LLM.py
@@ -100,94 +100,95 @@ if StateKeys.LLM_INPUT not in st.session_state:
 volcano_plot, plot_parameters = st.session_state[StateKeys.LLM_INPUT]
 
 st.markdown(f"Parameters used for analysis: `{plot_parameters}`")
-c1, c2 = st.columns((2, 1))
 
-with c2:
+c1, c2, c3 = st.columns((1, 1, 1))
+
+with c3:
+    st.markdown("##### Volcano plot")
     display_figure(volcano_plot.plot)
 
-with c1:
-    regulated_genes_df = volcano_plot.res[volcano_plot.res["label"] != ""]
-    regulated_genes_dict = dict(
-        zip(regulated_genes_df[Cols.INDEX], regulated_genes_df["color"].tolist())
-    )
+regulated_genes_df = volcano_plot.res[volcano_plot.res["label"] != ""]
+regulated_genes_dict = dict(
+    zip(regulated_genes_df[Cols.INDEX], regulated_genes_df["color"].tolist())
+)
 
-    if not regulated_genes_dict:
-        st.text("No genes of interest found.")
-        st.stop()
+if not regulated_genes_dict:
+    st.text("No genes of interest found.")
+    st.stop()
 
-    # Separate upregulated and downregulated genes
-    upregulated_genes = [
-        key for key in regulated_genes_dict if regulated_genes_dict[key] == "up"
-    ]
-    downregulated_genes = [
-        key for key in regulated_genes_dict if regulated_genes_dict[key] == "down"
-    ]
+# Separate upregulated and downregulated genes
+upregulated_genes = [
+    key for key in regulated_genes_dict if regulated_genes_dict[key] == "up"
+]
+downregulated_genes = [
+    key for key in regulated_genes_dict if regulated_genes_dict[key] == "down"
+]
 
-    # Create dataframes with checkboxes for selection
-    if st.session_state[StateKeys.SELECTED_GENES_UP] is None:
-        st.session_state[StateKeys.SELECTED_GENES_UP] = upregulated_genes
-    upregulated_genes_df = pd.DataFrame(
-        {
-            "Gene": [
-                st.session_state[StateKeys.DATASET]._feature_to_repr_map[protein]
-                for protein in upregulated_genes
-            ],
-            "Selected": [
-                protein in st.session_state[StateKeys.SELECTED_GENES_UP]
-                for protein in upregulated_genes
-            ],
-            "Protein": upregulated_genes,
-        }
-    )
-
-    if st.session_state[StateKeys.SELECTED_GENES_DOWN] is None:
-        st.session_state[StateKeys.SELECTED_GENES_DOWN] = downregulated_genes
-    downregulated_genes_df = pd.DataFrame(
-        {
-            "Gene": [
-                st.session_state[StateKeys.DATASET]._feature_to_repr_map[protein]
-                for protein in downregulated_genes
-            ],
-            "Selected": [
-                protein in st.session_state[StateKeys.SELECTED_GENES_DOWN]
-                for protein in downregulated_genes
-            ],
-            "Protein": downregulated_genes,
-        }
-    )
-
-    st.markdown("##### Genes of interest")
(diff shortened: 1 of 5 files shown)


alphastats/gui/pages/06_LLM.py
```diff
@@ -114,42 +114,91 @@ def llm_config():
...
+                    default=True,
+                ),
+                "Protein": st.column_config.TextColumn(
+                    "Protein",
+                    help="The protein name to be included in the analysis",
+                    width="medium",
+                ),
+            },
+            disabled=["Protein"],
+            hide_index=True,
```
mschwoer, line 178: please extract a method here to avoid code dup

alphastats/gui/utils/analysis_helper.py
```diff
@@ -202,14 +202,14 @@ def gather_parameters_and_do_analysis(
         raise ValueError(f"Analysis method {analysis_method} not found.")
 
 
-def gather_uniprot_data(features: list) -> None:
+def gather_uniprot_data(features: list[str]) -> None:
```
mschwoer: `List[str]` .. we're still at 3.8 :-)

alphastats/gui/pages/06_LLM.py
```diff
@@ -114,42 +114,91 @@ def llm_config():
...
+    # If no genes are selected, stop the script
+    if not regulated_genes_dict:
+        st.text("No genes selected for analysis.")
+        st.stop()
+
+    if c1.button("Gather UniProt data for selected proteins"):
+        gather_uniprot_data(selected_regulated_genes)
+
+if not st.session_state[StateKeys.ANNOTATION_STORE]:
+    st.stop()
```
mschwoer, line 201: maybe display an `st.info` "Download uniprot data first" or smth

alphastats/gui/pages/06_LLM.py
```diff
@@ -114,42 +114,91 @@ def llm_config():
...
     upregulated_genes = [
         key for key in regulated_genes_dict if regulated_genes_dict[key] == "up"
     ]
     downregulated_genes = [
         key for key in regulated_genes_dict if regulated_genes_dict[key] == "down"
     ]
 
+    # Create dataframes with checkboxes for selection
+    upregulated_genes_df = pd.DataFrame(
+        {"Protein": upregulated_genes, "Selected": [True] * len(upregulated_genes)}
```
mschwoer, line 127: This displays a different type of if than the volcano and the rest of the page 
image

check out `st.session_state[StateKeys.DATASET]._feature_to_repr_map`

mschwoer (commented):
I like that feature already :-)

alphastats/gui/pages/06_LLM.py
boopthesnoot: what's the consensus? do we use genes or proteins everywhere?

mschwoer: `DataSet._create_id_dicts` defines the current consensus on the representation

boopthesnoot: thanks!

alphastats/gui/pages/06_LLM.py
```diff
@@ -114,43 +115,55 @@ def llm_config():
...
+
+    # If no genes are selected, stop the script
+    if not regulated_genes_dict:
+        st.text("No genes selected for analysis.")
+        st.stop()
+
+    if c1.button("Gather UniProt data for selected proteins"):
+        gather_uniprot_data(selected_regulated_genes)
+
+if not st.session_state[StateKeys.ANNOTATION_STORE]:
```
JuliaS92, line 163: The annotation store always exists and might also be filled with previous analysis annotations. Instead check if any regulated_genes_dict.keys (i guess) are not in the store yet.

alphastats/gui/utils/llm_helper.py
```diff
@@ -11,6 +12,48 @@
...
+
+    Returns:
+        selected_proteins (List[str]): A list of selected proteins.
+    """
+    st.write(title)
+    df.insert(
+        0,
+        "Gene",
+        [
+            st.session_state[StateKeys.DATASET]._feature_to_repr_map[protein]
```
JuliaS92, line 30: Maybe pass the map as argument, so the funtion can be tested without a dummy dataset.

alphastats/gui/pages/06_LLM.py
JuliaS92: We use the feature internally and map it to the representation whenever needed in the ui. You solved it exaclty right :)

alphastats/gui/pages/06_LLM.py
JuliaS92: I changed this, but also removed the st.stop, as one can in principle move on without this.

mschwoer (approved):
LGTM

alphastats/gui/utils/llm_helper.py
```diff
@@ -11,6 +12,48 @@
 )
 
 
+def create_protein_editor(
```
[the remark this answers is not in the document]
JuliaS92: Following streamlit convention (a widgets return value is not the widget, but the value) I called it protein_selector now.

alphastats/gui/pages/06_LLM.py
```diff
@@ -124,35 +124,60 @@ def llm_config():
...
+            "Gene": [
+                st.session_state[StateKeys.DATASET]._feature_to_repr_map[protein]
+                for protein in upregulated_genes
+            ],
+            "Selected": [
+                protein in st.session_state[StateKeys.SELECTED_GENES_UP]
+                for protein in upregulated_genes
+            ],
+            "Protein": upregulated_genes,
+        }
```
mschwoer: could this be a small method? (code dup. right below)

alphastats/gui/pages/06_LLM.py
```diff
@@ -124,35 +124,60 @@ def llm_config():
...
     with c12:
-        selected_downregulated_genes = protein_selector(
+        st.session_state[StateKeys.SELECTED_GENES_DOWN] = protein_selector(
             downregulated_genes_df,
             "Downregulated Proteins",
-            st.session_state[StateKeys.DATASET]._feature_to_repr_map,
         )
 
     # Combine the selected genes into a new regulated_genes_dict
-    selected_regulated_genes = selected_upregulated_genes + selected_downregulated_genes
```
mschwoer: (nit) just `selected_genes`?

alphastats/gui/utils/ui_helper.py
```diff
@@ -151,6 +157,8 @@ class StateKeys(metaclass=ConstantsClass):
     LLM_INPUT = "llm_input"
     LLM_INTEGRATION = "llm_integration"
     ANNOTATION_STORE = "annotation_store"
+    SELECTED_GENES_UP = "selected_genes_up"
```
mschwoer, line 160: just a thought, we could use a naming convention here to distinguish "global" from "page-local" keys,
e.g. 
`P_SELECTED_GENES_UP = "selected_genes_up"`

alphastats/gui/utils/llm_helper.py
```diff
@@ -12,27 +12,24 @@
...
     st.write(title)
-    df = df.copy()
-    df.insert(
-        0,
-        "Gene",
-        [feature_to_repr_map[protein] for protein in df["Protein"]],
-    )
-    df.insert(2, "Protein", df.pop("Protein"))
+    if st.button("Select all", help=f"Select all {title} for analysis"):
+        st.session_state[StateKeys.SELECTED_GENES_UP] = df["Protein"].tolist()
```
mschwoer: we have this logic (fill with ALL) now at two places: here and in l.143
```
    if st.session_state[StateKeys.SELECTED_GENES_DOWN] is None:
        st.session_state[StateKeys.SELECTED_GENES_DOWN] = downregulated_genes
```
could we not just set it to `None` here?

alphastats/gui/utils/ui_helper.py
```diff
@@ -131,6 +131,12 @@ def init_session_state() -> None:
     if StateKeys.ANNOTATION_STORE not in st.session_state:
         st.session_state[StateKeys.ANNOTATION_STORE] = {}
 
+    if StateKeys.SELECTED_GENES_UP not in st.session_state:
+        st.session_state[StateKeys.SELECTED_GENES_UP] = None
+
+    if StateKeys.SELECTED_GENES_DOWN not in st.session_state:
+        st.session_state[StateKeys.SELECTED_GENES_DOWN] = None
+
```
mschwoer, line 139: just realizing this could be something like
```
defaults = {
StateKeys.USER_SESSION_ID: str(uuid.uuid4()),
StateKeys.ANALYSIS_LIST: []
..
}

for k, v in default.items():
       if k not in st.session_state:
        st.session_state[k] = v
```
(not now)

mschwoer (approved):

alphastats/gui/utils/llm_helper.py
JuliaS92: Yes, this would have the same effect, but they are two different cases (on llm initialization vs on select all) and this way is more explicit and the state does not rely on reruns of the app and the initialization from the page code. So I would leave it as is.

alphastats/gui/utils/ui_helper.py
JuliaS92: Agreed :)

alphastats/gui/utils/ui_helper.py
JuliaS92: board: https://github.com/orgs/MannLabs/projects/12/views/1?pane=issue&itemId=93996487

mschwoer: could also be just a TODO in the code ;-)

PR #402, opened 2025-03-11
move session state keys to dedicated module

mschwoer:
Move session state keys to dedicated module


alphastats/gui/utils/state_keys.py
```diff
@@ -0,0 +1,28 @@
+"""Keys and helper functions for the session state."""
```
JuliaS92, line 1: Why are the functions and the keys in two separate files now?
If that has a good reason, please remove the 'helper functions' from the docs here and explain instead where they are.

alphastats/gui/utils/llm_helper.py
```diff
@@ -4,7 +4,8 @@
 import pandas as pd
 import streamlit as st
 
-from alphastats.gui.utils.ui_helper import DefaultStates, StateKeys
+from alphastats.gui.utils.state_keys import StateKeys
+from alphastats.gui.utils.ui_helper import DefaultStates
```
JuliaS92, line 8: I we were to move the defaults as well it might make sense to create:
- alphastats.gui.utils.state.keys
- alphastats.gui.utils.state.defaults
- alphastats.gui.utils.state.utils

JuliaS92 (approved):
I added a few comments.
Biggest question: hat was the motivation for this change?

alphastats/gui/utils/state_keys.py
mschwoer: sorry for not mentioning this: the motivation was to resolve a circular import provlem in #403

alphastats/gui/utils/llm_helper.py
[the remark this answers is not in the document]
mschwoer: circular imports again :-)

mschwoer: if this further grows, might make sense one day.. for now, I'd like to keep it flat

PR #405, opened 2025-03-14
add version to pickle

mschwoer:
Add version to pkl

### DIFF
--- a/alphastats/gui/utils/session_manager.py
+++ b/alphastats/gui/utils/session_manager.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import logging
 from datetime import datetime
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


alphastats/gui/utils/session_manager.py
```diff
@@ -71,12 +81,17 @@ def save(self, session_state: SessionStateProxy, session_name: str = "") -> str:
 
         timestamp = datetime.now(tz=pytz.utc).strftime("%Y%m%d-%H%M%S")
         session_name = f"_{session_name}" if session_name else ""
-        file_name = f"{_PREFIX}_{timestamp}{session_name}_v{__version__}.{_EXT}"
+        file_name = f"{_PREFIX}_{timestamp}{session_name}.{_EXT}"
+
+        data_to_dump = {
+            SavedSessionKeys.STATE: state_data_to_save,
```
JuliaS92, line 87: I'm not sure adding this layer in the dict is necessary - I guess I would have added the version as a regular session_state key and used the previous mechanism.

However, with this mechanism we could also add the timestamp and session_name to the guts of the pickle as well.

Feel free to add or ticket this.

JuliaS92 (approved):

alphastats/gui/utils/session_manager.py
mschwoer: added .. I introduced this layer to not pollute our beloved session state :-)

PR #409, opened 2025-03-19
Improve lllm experience iii

mschwoer:
- display chat on results page
- some minor refactorings
- some UI tweaks (sorry got down the usual UI rabbit hole)

PR #411, opened 2025-03-19
Simplify and fix protein selector

JuliaS92:
Bug was that click None, select few, click None didn't work. Same for the all button. The second click didn't do anything.

Fix: check if list changes upon edit and st.rerun

Simplification: Only rerun the fragment - this required adding the df generation into the fragment.

Note: This makes is a necessity to always update the initial prompt (full page rerun would have avoided that), which is already the case for other changes.

### DIFF
--- a/alphastats/gui/pages/06_LLM.py
+++ b/alphastats/gui/pages/06_LLM.py
@@ -17,7 +17,6 @@ from alphastats.gui.utils.llm_helper import (
     LLM_ENABLED_ANALYSIS,
     OLLAMA_BASE_URL,
     display_uniprot,
-    get_df_for_protein_selector,
     protein_selector,
 )
 from alphastats.gui.utils.state_keys import DefaultStates, LLMKeys, StateKeys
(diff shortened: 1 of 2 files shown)


alphastats/gui/pages/06_LLM.py
```diff
@@ -128,21 +127,15 @@ def _pretty_print_analysis(key: str) -> str:
 # Create dataframes with checkboxes for selection
 if selected_llm_chat.get(LLMKeys.SELECTED_GENES_UP) is None:
     selected_llm_chat[LLMKeys.SELECTED_GENES_UP] = upregulated_genes
-upregulated_genes_df = get_df_for_protein_selector(
```
mschwoer, line 131: this might collide with https://github.com/MannLabs/alphapeptstats/pull/409 & https://github.com/MannLabs/alphapeptstats/pull/410

mschwoer (approved):

alphastats/gui/pages/06_LLM.py
JuliaS92: https://github.com/MannLabs/alphapeptstats/pull/411/commits/c4684e555383c733161af335097cffa59bc159aa

PR #412, opened 2025-03-19
Suggestion state loading

JuliaS92:
Instead of creating state keys on the fly, use the on_change callbacks to keep the llm chat entry and the session state in sync.


alphastats/gui/utils/state_keys.py
```diff
@@ -24,7 +24,9 @@ class StateKeys(metaclass=ConstantsClass):
     LLM_CHATS = "llm_chats"
 
     ANNOTATION_STORE = "annotation_store"
+    SELECED_ANALYSIS = "selected_analysis"
```
mschwoer, line 27: typo: `SELECED`

alphastats/gui/utils/llm_helper.py
```diff
@@ -364,3 +355,23 @@ def display_uniprot(
...
                 )
             )
+
+
+def on_select_fill_state():
+    selected_analysis = st.session_state[StateKeys.SAVED_ANALYSES].get(
+        st.session_state[StateKeys.SELECED_ANALYSIS], None
+    )
+    st.session_state[StateKeys.INCLUDE_UNIPROT] = selected_analysis.get(
+        LLMKeys.INCLUDE_UNIPROT, False
```
mschwoer, line 365: would call the individual methods very specific, e.g.
`on_change_include_uniprot_into_initial_prompt()` and `on_change_select_analysis_to_discuss()`

alphastats/gui/utils/llm_helper.py
```diff
@@ -303,23 +302,15 @@ def display_uniprot(
...
-            st.session_state[session_state_key] = False
 
         st.checkbox(
             "Integrate into initial prompt",
             help="If this is ticked and the initial prompt is updated, the Uniprot information will be included in the prompt and the instructions regarding uniprot will change to onl;y look up more information if explicitly asked to do so. Make sure that the total tokens are below the message limit of your LLM.",
-            key=get_uniprot_state_key(selected_analysis_key),
-            value=st.session_state[
-                get_uniprot_state_key(selected_analysis_key)
-            ],  # st.session_state.get(get_uniprot_state_key(selected_analysis_key), False),
+            key=StateKeys.INCLUDE_UNIPROT,
```
mschwoer, line 309: `INCLUDE_UNIPROT` -> `INCLUDE_UNIPROT_INTO_INITIAL_PROMPT`

alphastats/gui/utils/llm_helper.py
```diff
@@ -364,3 +355,23 @@ def display_uniprot(
...
+
+
+def on_change_save_state():
+    selected_analysis = st.session_state[StateKeys.SAVED_ANALYSES].get(
+        st.session_state[StateKeys.SELECED_ANALYSIS], None
+    )
+    selected_analysis[LLMKeys.INCLUDE_UNIPROT] = st.session_state[
+        StateKeys.INCLUDE_UNIPROT
+    ]
+    st.toast("State saved.", icon="💾")
```
mschwoer, line 377: this toast is a bit confusing to non-developers :-) would drop it

mschwoer (approved):
we should use more callbacks I guess ;-)

could collide with https://github.com/MannLabs/alphapeptstats/pull/410 though

alphastats/gui/utils/llm_helper.py
JuliaS92: Yes, this was just included to demonstrate the behaviour.

alphastats/gui/utils/llm_helper.py
JuliaS92: I was thinking more in the direction of one callback to load them all and one callback to save them all, as they are all small values and saving doesn't trigger anything. That way we can just use the same callback on all widgets related to this.

mschwoer: also a valid argument :-)

JuliaS92: I added doc strings to the two functions to make their intention clear.

PR #417, opened 2025-03-26
Fix intensity plot

JuliaS92:
The handling of requested ids was very fragile with different LLM inputs.

The new function maps the best matching entry and can also deal with lists.

Some reformatting of the intensity plot included:
- display representation instead of protein id
- adjust width automatically based on number of proteins and groups

### DIFF
--- a/alphastats/dataset/dataset.py
+++ b/alphastats/dataset/dataset.py
@@ -487,23 +487,6 @@ class DataSet:
 
         return volcano_plot.plot
 
-    def _get_features_for_gene_name(
-        self,
-        gene_name: str,
-    ) -> list:
-        """Get feature from gene name. If gene name is not present, return gene name, as we might already have a gene id.
-        'HEL114' -> ['P18206;A0A024QZN4;V9HWK2;B3KXA2;Q5JQ13;B4DKC9;B4DTM7;A0A096LPE1']
-
-        Args:
-            gene_name (str): Gene name
-
-        Returns:
-            list: Protein group ids or gene name if not present in the mapping.
-        """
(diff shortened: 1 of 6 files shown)


alphastats/llm/llm_functions.py
```diff
@@ -198,10 +198,9 @@ def get_assistant_functions(
                 "parameters": {
                     "type": "object",
                     "properties": {
-                        "gene_name": {
+                        "feature": {
                             "type": "string",
-                            "enum": genes_of_interest,
-                            "description": "Identifier for the gene of interest",
+                            "description": "Identifier for the feature of interest, use the same format as in the initial prompt, inidivual gene symbols, or individual protein ids",
```
mschwoer, line 203: `Identifier for the feature of interest. Use the same format as in the initial prompt: inidividual gene symbols, or individual protein ids.`

alphastats/dataset/dataset.py
```diff
@@ -487,28 +488,60 @@ def plot_volcano(
 
         return volcano_plot.plot
 
-    def _get_features_for_gene_name(
-        self,
-        gene_name: str,
-    ) -> list:
-        """Get feature from gene name. If gene name is not present, return gene name, as we might already have a gene id.
-        'HEL114' -> ['P18206;A0A024QZN4;V9HWK2;B3KXA2;Q5JQ13;B4DKC9;B4DTM7;A0A096LPE1']
+    def _get_feature_ids_from_string(self, string: str) -> List[str]:
```
mschwoer, line 491: adding tests via LLM should be a breeze for these two

mschwoer (approved):

PR #418, opened 2025-03-26
Accurate token count

boopthesnoot:
getting the token count info from the API call if available, with token estimation as a fallback


JuliaS92 (approved):
Thank you for fixing the token count. I think we need to do something about the tool tokens though - this makes few-shot engineering in tool descriptions very expensive.

JuliaS92:
Sorry I just realized one point late: Does this new counting affect the history truncation as well, or do you think we should just take the tool tokens as a given on top of the 'visible' history?

boopthesnoot:
> Sorry I just realized one point late: Does this new counting affect the history truncation as well, or do you think we should just take the tool tokens as a given on top of the 'visible' history?

great point, the truncation should be based on real API tokens, I'll check it tomorrow

alphastats/llm/llm_integration.py
```diff
@@ -618,3 +648,13 @@ def _display_artifact(self, artifact):
             display(HTML(pio.to_html(artifact, full_html=False)))
         else:
             display(Markdown(f"```\n{str(artifact)}\n```"))
+
+    def get_token_usage(self) -> Dict[str, int]:
```
mschwoer, line 652: could this be a `@property`?

alphastats/llm/llm_integration.py
```diff
@@ -488,6 +516,8 @@ def get_print_view(
                 }
             )
 
+        total_tokens += tools_tokens * api_call_count
```
mschwoer, line 519: not sure if I get this logic? do we send all tolls with every api call?

alphastats/llm/llm_integration.py
```diff
@@ -463,7 +478,20 @@ def get_print_view(
...
         pinned_tokens = 0
+        api_call_count = 0
+
+        tools_tokens = 0
+        if self._tools:
+            tools_str = json.dumps(self._tools)
+            tools_tokens = self.estimate_tokens(
+                [{MessageKeys.ROLE: "system", MessageKeys.CONTENT: tools_str}],
+                self._model,
+            )
```
mschwoer, line 489: this is constant for a session, right? could this be moved to where the tools are initialized?

alphastats/llm/llm_integration.py
JuliaS92: Yes, that's exactly what's happening. It is currently not a lot, but hence my comment above, that this makes extensive prompt engineering with examples "expensive".

boopthesnoot: i assume that this allows us to dynamically change the available tools, but it's just my understanding, not sure about how this all works, maybe I'm wrong big time

mschwoer: @boopthesnoot dynamically, you mean during runtime? this I would do only if this was really expensive..
however,  we should add this as a comment, e.g. to the place where the tools are defined
"Note: we send all tools with every api call, so every tool increases token count."

alphastats/llm/llm_integration.py
boopthesnoot: yes

JuliaS92:
I had issues running this, but it might be model specific. I'll let you know once I found out the issue.

alphastats/gui/utils/llm_helper.py
```diff
@@ -903,9 +903,15 @@ def show_llm_chat(
                     st.warning("Don't know how to display artifact:")
                     st.write(artifact)
 
-    st.markdown(
-        f"*total tokens in context: {str(total_tokens)}, tokens used for pinned messages: {str(pinned_tokens)}*"
-    )
+    token_usage = llm_integration.get_token_usage
+    if token_usage["total_tokens"] > 0:
+        st.markdown(
+            f"*Tokens used: {token_usage['total_tokens']} (prompt: {token_usage['prompt_tokens']}, completion: {token_usage['completion_tokens']}), pinned messages: {str(pinned_tokens)}*"
```
JuliaS92, line 909: Can we somehow make it clear, that these are all the tokens used throughout the conversation and not, how large the current context window is?
If we could somehow determine the context window size for the next completion as well that would be great (as this is what we estimate for the truncation), but this can be done separately.

JuliaS92 (approved):
We still need to refine how this interacts with the history truncation, but this is a viable slice.

PR #421, opened 2025-03-31
Modular initial prompt

JuliaS92:
To facilitate prompt engineering, the initial prompt is split into several text areas, that can be reset and prefilled individually and are saved with the chat to be persistent.

image

### DIFF
--- a/alphastats/gui/utils/llm_helper.py
+++ b/alphastats/gui/utils/llm_helper.py
@@ -1,7 +1,9 @@
+from __future__ import annotations
+
 import os
 import warnings
 from pathlib import Path
-from typing import List, Optional
+from typing import Optional
 
 import pandas as pd
 import streamlit as st
@@ -191,7 +193,7 @@ def protein_selector(
 
 
 def get_df_for_protein_selector(
-    proteins: List[str], selected: List[str]
+    proteins: list[str], selected: list[str]
 ) -> pd.DataFrame:
     """Create a DataFrame for the protein selector.
 
@@ -214,7 +216,7 @@ def get_df_for_protein_selector(
 
 
 def get_display_proteins_html(
-    protein_ids: List[str], is_upregulated: True, annotation_store, feature_to_repr_map
+    protein_ids: list[str], is_upregulated: True, annotation_store, feature_to_repr_map
 ) -> str:
     """
     Get HTML code for displaying a list of proteins, color according to expression.
@@ -273,9 +275,9 @@ def set_api_key(api_key: str = None) -> None:
 
 def llm_connection_test(
     model_name: str,
-    base_url: Optional[str] = None,
-    api_key: Optional[str] = None,
-) -> Optional[str]:
+    base_url: Optional | str = None,
+    api_key: Optional | str = None,
+) -> Optional | str:
     """Test the connection to the LLM API, return None in case of success, error message otherwise."""
     try:
         llm = LLMIntegration(
--- a/alphastats/gui/utils/llm_helper.py
+++ b/alphastats/gui/utils/llm_helper.py
@@ -18,6 +18,7 @@ from alphastats.gui.utils.state_keys import (
 )
 from alphastats.llm.llm_integration import LLMIntegration, MessageKeys, Models, Roles
 from alphastats.llm.prompts import (
+    LLMInstructionKeys,
     _get_experimental_design_prompt,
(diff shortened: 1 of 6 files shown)


alphastats/gui/utils/llm_helper.py
```diff
@@ -108,26 +121,70 @@ def init_llm_chat_state(
...
+                llm_chat[LLMKeys.SELECTED_GENES_UP],
+            )
+        ),
+        list(
+            map(
+                feature_to_repr_map.get,
+                llm_chat[LLMKeys.SELECTED_GENES_DOWN],
+            )
+        ),
+        uniprot_info,
```
mschwoer, line 177: could this be moved to `_get_protein_data_prompt`?

alphastats/gui/utils/llm_helper.py
```diff
@@ -423,12 +483,25 @@ def display_uniprot(
...
-        selected_analysis.get(LLMKeys.INCLUDE_UNIPROT_INTO_INITIAL_PROMPT, False)
+    st.session_state[StateKeys.PROMPT_EXPERIMENTAL_DESIGN] = selected_chat.get(
+        LLMKeys.PROMPT_EXPERIMENTAL_DESIGN, None
     )
+    st.session_state[StateKeys.PROMPT_PROTEIN_DATA] = selected_chat.get(
+        LLMKeys.PROMPT_PROTEIN_DATA, None
+    )
+    st.session_state[StateKeys.PROMPT_INSTRUCTIONS] = selected_chat.get(
+        LLMKeys.PROMPT_INSTRUCTIONS, None
+    )
```
mschwoer, line 500: just noting that my idea of double-bookkeeping is getting out of hand quite quickly :-)

a idea for a future refactoring could be to store these relations centrally
```
map = [
StateKeys.PROMPT_EXPERIMENTAL_DESIGN: (LLMKeys.PROMPT_EXPERIMENTAL_DESIGN, None)
,
...
]
```
and then do this mapping and the reverse one (l. 520) programatically

alphastats/llm/prompts.py
```diff
@@ -48,17 +59,47 @@ def get_initial_prompt(
...
+    SIMPLE = "simple"
+    CUSTOM = "customize prompt"
+
+
+LLMInstructions = {
+    LLMInstructionKeys.SIMPLE: (
+        "Help us understand the potential connections between these proteins and how they might be contributing "
+        "to the differences. After that provide a high level summary."
+    ),
+    LLMInstructionKeys.CUSTOM: "Please give instructions to the LLM model on how to generate a response.",
```
mschwoer, line 80: `<<<Please give instructions to the LLM model on how to generate a response.>>>` ?

mschwoer (approved):

alphastats/gui/utils/llm_helper.py
JuliaS92: That was a suggestion from Claude as well. To handle the model parameters more easily I would actually suggest to move the config back to this page, as we can then just work with disableing it instead of checking if the model is aready initialized.

JuliaS92: Decision after offline discussion: Make a mapping for the ones independent of initialization and keep separate logic for Model related parameters, to be retain ability to move them around.

JuliaS92: https://github.com/MannLabs/alphapeptstats/pull/423

alphastats/gui/utils/state_utils.py
```diff
@@ -15,33 +15,26 @@ def empty_session_state() -> None:
...
-            Models.GPT4O
-        )  # TDOO: change to None: this is just for convenience now
-
-    if StateKeys.SELECTED_ANALYSIS not in st.session_state:
-        st.session_state[StateKeys.SELECTED_ANALYSIS] = None
+def init_session_state() -> None:
+    """Initialize the session state if not done yet."""
+    for key, value in INIT_STATES.items():
+        if key not in st.session_state:
+            st.session_state[key] = value
```
mschwoer, line 40: careful with mutable args here!
```
INIT_STATES = {
    "bla":[]
}

something = INIT_STATES["bla"]
something.append(1)

print(INIT_STATES["bla"])
```
need to make a `.copy()` for non-None

mschwoer (approved):
LGTM but add the `.copy()`

PR #425, opened 2025-04-07
add deselect widget to protein selector

JuliaS92:
This is so we can reproduce papers exactly. Make square cutoffs that include all significant proteins, then deselct the ones, that were not reported significant in the original paper.

### DIFF
--- a/alphastats/gui/utils/llm_helper.py
+++ b/alphastats/gui/utils/llm_helper.py
@@ -209,6 +209,19 @@ def protein_selector(
     if c2.button("Select none", help=f"Select no {title} for analysis"):
         selected_analysis_session_state[state_key] = []
         st.rerun(scope="fragment")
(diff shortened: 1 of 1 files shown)


alphastats/gui/utils/llm_helper.py
```diff
@@ -209,6 +209,19 @@ def protein_selector(
     if c2.button("Select none", help=f"Select no {title} for analysis"):
         selected_analysis_session_state[state_key] = []
         st.rerun(scope="fragment")
+    deselect = st.text_input(
```
mschwoer, line 212: mb add a comment
`# This is so we can reproduce papers exactly. Make square cutoffs that include all significant proteins, then deselct the ones, that were not reported significant in the original paper.`


.. and a `help`?

mschwoer (approved):

PR #427, opened 2025-04-16
Uniprot fetching

JuliaS92:
Lock uniprot info fragment on LLM initialization.
Add uniprot fields as parameter to llm tool and default to all fields.

In the process:
- deduplicate code for selecting features based on search string
- update type hints

### DIFF
--- a/alphastats/llm/llm_functions.py
+++ b/alphastats/llm/llm_functions.py
@@ -1,6 +1,8 @@
 """Module for defining assistant functions for the ChatGPT model."""
 
-from typing import Dict, List, Tuple, Union
+from __future__ import annotations
+
+from typing import Optional
 
 import pandas as pd
 import streamlit as st
@@ -9,6 +11,7 @@ from alphastats.dataset.dataset import DataSet
 from alphastats.gui.utils.state_keys import StateKeys
 from alphastats.llm.enrichment_analysis import get_enrichment_data, gprofiler_organisms
 from alphastats.llm.uniprot_utils import (
+    ExtractedUniprotFields,
     format_uniprot_annotation,
     get_annotations_for_feature,
 )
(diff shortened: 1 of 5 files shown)


alphastats/llm/llm_functions.py
```diff
@@ -137,6 +127,11 @@ def get_general_assistant_functions() -> List[Dict]:
                             "type": "string",
                             "description": "Feature representation, gene identifier or protein id",
                         },
+                        "fields": {
```
mschwoer, line 130: we did not need this before?

alphastats/llm/llm_functions.py
```diff
@@ -71,43 +76,28 @@ def get_uniprot_info_for_search_string(search_string: str) -> str:
...
+    if not fields:
+        fields = list(annotation.keys())
+
     return (
         search_string
         + ": "
         + format_uniprot_annotation(
             annotation,
-            fields=st.session_state[StateKeys.SELECTED_UNIPROT_FIELDS],
+            fields=fields,
```
mschwoer, line 100: `fields=fields if fields is not None else list(annotation.keys())` ?

alphastats/llm/llm_functions.py
```diff
@@ -71,43 +76,28 @@ def get_uniprot_info_for_search_string(search_string: str) -> str:
...
-
-    elif search_string in gene_to_features_map:
-        features = gene_to_features_map[search_string]
-
-    elif search_string in protein_to_features_map:
-        features = protein_to_features_map[search_string]
-    else:
-        raise ValueError(f"{search_string} not found in dataset.")
+    try:
+        features = dataset._get_feature_ids_from_string(search_string)
```
mschwoer, line 83: could we rename this method to not be confused with the "string db"?

mschwoer (approved):

alphastats/llm/llm_functions.py
JuliaS92: before it would use the fields from the widgets.

PR #433, opened 2025-05-06
Switch to Pyproject.toml

JuliaS92:
This is to get rid of setup.py based installation and to fix any issues with undetected submodules.


mschwoer (approved):
LGTM (please clean up `__init__.py`)

PR #437, opened 2025-05-16
Bumpversion

mschwoer:
- add bumpversion workflow
- repair the release
see it in action here https://github.com/MannLabs/alphapeptstats/actions/runs/15073444355 and the draft release here: https://github.com/MannLabs/alphapeptstats/releases/tag/untagged-6d4ce4f40b2b50277f88 
Checked the Mac ARM an windows installers


alphastats/__init__.py
```diff
@@ -1 +1 @@
-__version__ = "0.6.9"
+__version__ = "0.6.10-dev0"
```
JuliaS92, line 1: I would call this at least 0.7

release/pyinstaller/alphastats.spec
```diff
@@ -21,85 +16,33 @@ else:
 block_cipher = None
 location = os.getcwd()
 project = "alphastats"
-remove_tests = True
```
JuliaS92, line 24: why not remove tests?

JuliaS92 (approved):
Looks like a lot of magic to me, but it works :)

alphastats/__init__.py
mschwoer: yes, this is just the current development version (which assumes that a bug fix is the next release)..
from there, one decides before the release on the release version

release/pyinstaller/alphastats.spec
mschwoer: did no do anything here

PR #440, opened 2025-05-21
Hide old dea

JuliaS92:
This is just to remove them before we release, as they don't perform MHC and should therefore not be used.
Cleaning up the codebase is a major task and will involve salvaging and merging some functionality into the new DEA.

### DIFF
--- a/alphastats/gui/utils/analysis.py
+++ b/alphastats/gui/utils/analysis.py
@@ -597,7 +597,10 @@ class DifferentialExpressionTwoGroupsAnalysis(AbstractGroupCompareAnalysis):
         ), None  # None is for backwards compatibility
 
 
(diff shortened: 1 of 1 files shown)


alphastats/gui/utils/analysis.py
```diff
@@ -598,15 +598,13 @@ def _do_analysis(self) -> Tuple[ResultComponent, None]:
 
 
 ANALYSIS_OPTIONS = {
-    PlottingOptions.VOLCANO_PLOT: VolcanoPlotAnalysis,
```
mschwoer, line 601: could you comment them out only, and add a `TODO: merge with new DEA` ..
maybe already create a ticket describing what is to be transferred

mschwoer (approved):

PR #449, opened 2025-06-04
do not send custom fields to LLM

mschwoer:
Do not send custom fields to LLM to avoid spurious error messages.

### DIFF
--- a/alphastats/llm/llm_integration.py
+++ b/alphastats/llm/llm_integration.py
@@ -95,6 +95,10 @@ class Model:
         return list(Model.MODELS.keys())
 
 
+# we do not want to pass our internal keys to the LLM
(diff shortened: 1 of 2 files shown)


alphastats/gui/utils/llm_helper.py
```diff
@@ -921,13 +921,6 @@ def show_llm_chat(
 
         st.rerun(scope="fragment")
 
-    st.download_button(
```
JuliaS92, line 924: Please add a back to top link here, so the button is still easily reachable:
st.markdown('[Back to Top](#llm-interpretation-with-openai-gpt-4o)')
Just need to add model formatting

JuliaS92 (approved):

PR #452, opened 2025-06-08
Allow custom analysis

mschwoer:


alphastats/gui/pages_/06_LLM.py
```diff
@@ -38,23 +40,24 @@
...
         key
         for key, analysis in st.session_state[StateKeys.SAVED_ANALYSES].items()
         if analysis[SavedAnalysisKeys.METHOD] in LLM_ENABLED_ANALYSIS
     ]
+    + [CUSTOM_ANALYSIS_KEY]
 ):
+    if dataset is None:
+        st.info("Import data first.")
+        st.stop()
+
```
JuliaS92, line 60: I wouldn't put this here, but instead have a custom analysis upload on the results page and then it shows up as a regular dropdown entry here.

alphastats/gui/pages_/06_LLM.py
```diff
@@ -70,23 +73,67 @@
...
+    volcano_plot = None
+    plot_parameters = fake_plot_parameters
+
+    regulated_features_dict = fake_regulated_features_dict
+    subgroups = fake_subgroups
+
+    *_, feature_to_repr_map = DataSet.create_id_dicts(
+        list(fake_regulated_features_dict.keys()),
+        list(fake_regulated_features_dict.keys()),
+    )  # TODO where to get this from?
```
JuliaS92, line 97: On loading the result, the mapping should be created for all available ids, not just the regulated ones.

alphastats/gui/pages_/06_LLM.py
```diff
@@ -95,24 +142,16 @@
 
 ##################################### Volcano plot #####################################
 
-with c3:
-    st.markdown("##### Volcano plot")
-    display_figure(volcano_plot.plot)
+if dataset:
+    with c3:
+        st.markdown("##### Volcano plot")
+        display_figure(volcano_plot.plot)
```
JuliaS92, line 148: Ideally we don't even have to disable this, but read the p-value and fold change if they are available.

JuliaS92 (commented):
Looks like a good start :)

alphastats/gui/pages_/06_LLM.py
mschwoer: I see, the LLM page should not know where the analysis comes from .. => will be another PR anyway

alphastats/gui/pages_/06_LLM.py
mschwoer: you mean we should show the volcano plot also for custom analysis?

JuliaS92: https://github.com/orgs/MannLabs/projects/12/views/1?filterQuery=component%3Aalphapeptstats&pane=issue&itemId=115500611

PR #459, opened 2025-06-20
Gemini models

JuliaS92:
I added a mechanism to use vertex_ai gemini models.
Gemini was the first to validate the tools schema, so that was fixed now, too.

### DIFF
--- a/alphastats/gui/utils/llm_helper.py
+++ b/alphastats/gui/utils/llm_helper.py
@@ -152,6 +152,8 @@ def llm_config() -> None:
 
 def format_analysis_key(key: str) -> str:
     """Pretty print an analysis referenced by `key`."""
+    if key not in st.session_state[StateKeys.SAVED_ANALYSES]:
+        return key
     analysis = st.session_state[StateKeys.SAVED_ANALYSES][key]
     return f"[{key}] #{analysis[SavedAnalysisKeys.NUMBER]} {analysis[SavedAnalysisKeys.METHOD]} {analysis[SavedAnalysisKeys.PARAMETERS]}"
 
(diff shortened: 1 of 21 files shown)


alphastats/llm/llm_integration.py
```diff
@@ -150,19 +155,51 @@ def __init__(
...
+    ) -> Dict[str, None]:
+        """Configure the model with the given parameters.
+
+        model_name : str
+            The type of API to use, will be forwarded to the client.
+        base_url : str, optional
+            The base URL for the API, by default None. For vertex ai this is converted to the location.
+        api_key : str, optional
+            The API key for authentication, by default None. For vertex ai this is converted to the project.
+        """
```
mschwoer, line 183: The docstring should reflect that all keys need to be valid parameters of `completion()` 
(not happy with this solution anyway, but I have nothing better in mind right now)

alphastats/llm/llm_integration.py
```diff
@@ -150,19 +155,51 @@ def __init__(
...
 
-        self.api_key = api_key
-
-        self.base_url = base_url
+        self.model_config = self.configure_model(
+            model_name, base_url=base_url, api_key=api_key
+        )
         self.is_multimodal = model.is_multimodal()
 
+    def configure_model(
```
mschwoer, line 168: private method?

alphastats/gui/utils/llm_helper.py
```diff
@@ -93,16 +93,20 @@ def llm_config() -> None:
         supports_base_url = Model(new_model_name).supports_base_url()
```
mschwoer, line 94: `is_vertex_model = new_model_name.startswith("vertex")`, then reuse

mschwoer (approved):

PR #464, opened 2025-06-26
Hotfixes

JuliaS92:
- switched groups
- tool parameter defaults and examples
- mismatched ids in dataset and custom analysis

### DIFF
--- a/alphastats/gui/pages_/08_Upload Custom Analysis.py
+++ b/alphastats/gui/pages_/08_Upload Custom Analysis.py
@@ -70,13 +70,13 @@ if uploaded_file is not None:
     with param_col1:
         group1 = st.text_input(
             "Group 1",
-            placeholder="e.g., Control (negative fold change values mean a protein is upregulated in this group)",
+            placeholder="e.g., Treatment (positive fold change values mean a protein is upregulated in this group)",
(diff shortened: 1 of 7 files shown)


alphastats/llm/prompts.py
```diff
@@ -87,6 +87,10 @@ def _get_protein_data_prompt(
             downregulated_features,
         )
     )
+    if any(gene is None for gene in upregulated_genes + downregulated_genes):
```
mschwoer, line 90: could this be moved up to where the custom analysis is loaded?

alphastats/gui/pages_/08_Upload Custom Analysis.py
```diff
@@ -70,13 +70,13 @@
     with param_col1:
         group1 = st.text_input(
             "Group 1",
```
mschwoer, line 72: so we assume that "Group 1" is always "positive -> upregulated"? is this documented somewhere?

JuliaS92: This is hard coded in the way the parameters dictionary is parsed for the prompt and the standard in the field, because you alwyas calculate group1 - group2 in log space. I'll add another comment.

PR #468, opened 2025-07-17
Annotation search

JuliaS92:
LLM tool to search through the annotation store with a regex or a fixed string.


alphastats/llm/llm_functions.py
```diff
@@ -166,6 +211,38 @@ def get_general_assistant_functions() -> list[dict]:
                 },
             },
         },
+        {
+            "type": "function",
+            "function": {
+                "name": get_feature_list_by_annotation_search_string.__name__,
+                "description": "Get a list of feature representations matching a search string from based on the annotation store filled from UniProt. Use this for example when asked for all proteins sharing a specific characteristic. Make sure to represent all of the matching features in your interpretation.",
```
PatriciaSkowronek, line 218: An alternative phrasing could be:
If the user is asking for all proteins sharing a specific characteristic
**Action:** Invoke this [name] tool.
**Input to Tool:**...
**Expected Output from Tool:** List of feature representations matching a search string based on the annotation store filled from UniProt. Make sure to represent all of the matching features in your interpretation.

alphastats/llm/llm_functions.py
```diff
@@ -99,9 +100,53 @@ def get_uniprot_info_for_search_string(
...
 
 
+def get_feature_list_by_annotation_search_string(
+    search_string: str,
+    fields: Optional | list = None,
+    *,
+    use_regex: bool = False,
+):
+    """Get a list of features based on a search string and fields.
+    This is used to get the features that match the search string and fields from the annotation store.
```
mschwoer, line 110: nowadays I tend to ask myself: would an LLM get from the docstring what is happening here without looking at the code..
In this case, it could be more specific

alphastats/llm/llm_functions.py
```diff
@@ -99,9 +100,53 @@ def get_uniprot_info_for_search_string(
...
+
+    Args:
+        search_string (str): The search string to use to find features.
+        fields (Optional | list, optional): The fields to use to filter the features. Defaults to None.
+        use_regex (bool, optional): Whether to use regex for searching. Defaults to False.
```
mschwoer, line 115: describe here also the default search method (lowercase 'contains')

alphastats/llm/llm_functions.py
```diff
@@ -99,9 +100,53 @@ def get_uniprot_info_for_search_string(
...
+    fields: Optional | list = None,
+    *,
+    use_regex: bool = False,
+):
+    """Get a list of features based on a search string and fields.
+    This is used to get the features that match the search string and fields from the annotation store.
+
+    Args:
+        search_string (str): The search string to use to find features.
+        fields (Optional | list, optional): The fields to use to filter the features. Defaults to None.
```
mschwoer, line 114: Defaults to None, which means ... ?

alphastats/llm/llm_functions.py
mschwoer, line 218: (no action needed) idea to reduce redundancy:
Put the general docstring (without parameters) into a variable right before the method,
and reference it here (and add it to the docstring of the method using `method_name.__doc__ += ` or smth)

tests/llm/test_llm_functions.py
```diff
@@ -240,3 +241,70 @@ def test_get_uniprot_info_for_search_string(
     assert call_arg[0] in mock_session_state[StateKeys.ANNOTATION_STORE]
     assert result == expected_result
     dataset.id_holder.get_feature_ids_from_search_string.assert_called_with(llm_input)
+
+
+@patch("alphastats.llm.llm_functions.format_uniprot_annotation")
```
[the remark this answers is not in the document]
mschwoer: Classic LLM :-D

tests/llm/test_llm_functions.py
```diff
@@ -240,3 +241,70 @@ def test_get_uniprot_info_for_search_string(
...
+        f"{key}: {annotation[key]}" for key in fields if key in annotation
+    )
+
+    # Test case 1: Search without regex
+    search_string = "value1"
+    fields = ["field1"]
+    result = get_feature_list_by_annotation_search_string(search_string, fields)
+    assert result == "repr1"
+
+    # Test case 2: Search with regex
```
mschwoer, line 277: (nit) this could be split into several tests

mschwoer (approved):

alphastats/llm/llm_functions.py
JuliaS92: I tend to disagree, because the function description for an LLM, vs a human docstring reader might differ, e.g. including specific instructions when to trigger this.

mschwoer: Fully agree that the LLM needs extra information (which would be put into this dictionary). But the general docstring should be valid for both humans and llm(an)s

